PR #236, opened 2014-08-14
Working implementation of image grid for team media!

nwalters512:
This is by no means ready to merge.


nwalters512:
This should probably be redone with RecyclerView with the GridLayoutManager (or whatever it's called).

Issue #330, opened 2015-02-22
Use Reactive Elements


phil-lopreiato:
Next framework update: try and use some RxAndroid elements to simplify the background tasks. There appears to be a substantial benefit over using the regular AsyncTask framework. 

http://stablekernel.com/blog/replace-asynctask-asynctaskloader-rx-observable-rxjava-android-patterns/

https://github.com/ReactiveX/RxAndroid

PR #367, opened 2015-03-29
Retrofit Support

phil-lopreiato:
[Retrofit](https://github.com/square/retrofit) is a library that converts a REST API -> Java interface. We can simplify a lot of our HTTP code this way, I bet.

It can also return observables (see #330), so we can do event-driven processing of returned data


phil-lopreiato:
Probably not the way we want to go. There are better approaches

PR #379, opened 2015-04-05
Notification titles

1fish2:
This PR makes notification titles more informative, esp. on the lock screen and when a single notification summarizes multiple notifications.

I hope y'all like this as much as I do. The push notification is a wonderful feature during competition weekends. (I have more improvement ideas.)

This PR also adds an easy way to test notifications locally without setting up a server or modifying the app: Python script test_notification.py uses "adb shell am broadcast" to send an Intent to the app's CGM Intent receiver, and a small string resource lets debug builds receive those Intents. Now you can easily send simulated push notifications and test how the app handles them no matter what activity the app is currently showing, or if it's not currently visible, or even if the screen is locked.

This PR also fixes some crashes and NPEs bugs found while testing notifications.

This PR isn't short but it should be easy to read one commit at a time.


android/src/main/java/com/thebluealliance/androidclient/helpers/MatchHelper.java
```diff
@@ -95,7 +96,7 @@ public static TYPE fromKey(String key) {
         SHORT_TYPES.put(MatchHelper.TYPE.SEMI, "sf");
         SHORT_TYPES.put(MatchHelper.TYPE.FINAL, "f");
 
-        LONG_TYPES = new HashMap<>();
+        LONG_TYPES = new HashMap<>(); // TODO: I18N
```
phil-lopreiato: What does that TODO mean?

android/src/main/java/com/thebluealliance/androidclient/helpers/MatchHelper.java
```diff
@@ -593,14 +609,24 @@ public static String getMatchTitleFromMatchKey(String matchKey) {
             }
 
             if(set == null) {
-                // No set specified; this is a quals match
-                return LONG_TYPES.get(TYPE.fromShortType(type)) + " " + number;
+                // No set specified; this is a match like "Quals 10" (abbrev "Q10")
+                String format = abbrev ? "%1$s%2$s" : "%1$s %2$s";
```
phil-lopreiato: Maybe move these strings to a resources file

android/src/main/res/values/strings.xml
```diff
@@ -21,6 +21,7 @@
     <string name="year_selector_title_events">%1$s Events</string>
     <string name="year_selector_title_districts">%1$s Districts</string>
     <string name="select_year">Select Year</string>
+    <string name="cgm_sender_permission">com.google.android.c2dm.permission.SEND</string>
```
phil-lopreiato: cgm is a typo

test_notification.py
```diff
@@ -0,0 +1,187 @@
+#! /usr/bin/env python
```
phil-lopreiato: Love this. Great idea!

upcoming_match.py
```diff
@@ -0,0 +1,16 @@
+#! /usr/bin/env python
```
phil-lopreiato: Is this file redundant now?

android/src/main/AndroidManifest.xml
```diff
@@ -199,7 +199,7 @@
         </receiver>
         <receiver
             android:name="com.thebluealliance.androidclient.gcm.GCMBroadcastReceiver"
-            android:permission="com.google.android.c2dm.permission.SEND">
+            android:permission="@string/cgm_sender_permission">
```
phil-lopreiato: Probably also a typo

android/src/debug/res/values/strings.xml
```diff
@@ -4,4 +4,7 @@
     <string name="package_name">com.thebluealliance.androidclient.development</string>
     <string name="account_type">com.thebluealliance.androidclient.development.account.TBAACCOUNT</string>
 
+    <!-- An empty string in debug builds lets 'adb shell am broadcast' send intents to GCMBroadcastReceiver. -->
+    <string name="cgm_sender_permission"></string>
```
phil-lopreiato: Also a typo?

android/src/main/java/com/thebluealliance/androidclient/datafeed/deserializers/EventDeserializer.java
```diff
@@ -16,6 +16,14 @@
 
 public class EventDeserializer implements JsonDeserializer<Event> {
 
+    /**
+     * Returns true if the given element is null or JsonNull. This is handy for checking the result
+     * of JsonObject#get(), which is null if the requested key is absent.
+     */
+    static boolean isNull(JsonElement element) {
```
phil-lopreiato: This helper function  might be good to break out somewhere else and use in the other deserializers, as well

android/src/main/java/com/thebluealliance/androidclient/fragments/tasks/UpdateUserModelSettingsTaskFragment.java
```diff
@@ -17,11 +17,13 @@
...
 
     public UpdateUserModelSettingsTaskFragment() {
-
+        settings = new ModelNotificationFavoriteSettings();
     }
 
     public UpdateUserModelSettingsTaskFragment(ModelNotificationFavoriteSettings settings) {
         this.settings = settings;
+        // TODO(jerry): Call Fragment#setArguments(Bundle) to stash settings so the 0-arg
+        // constructor can call getArguments() to get them?
```
phil-lopreiato: That's the "correct" way to do it. If you use any of the fragment helper classes (or rotate the device), the fragment will be recreated with the 0-arg constructor. Probably a good idea to do

phil-lopreiato:
Looks pretty good! I really like the python script addition. Let's aim to get this in the next release (for champs)

android/src/main/java/com/thebluealliance/androidclient/helpers/MatchHelper.java
1fish2: "TODO" marks a searchable to-do comment.

IntelliJ's TODO tool window makes it easy to go through them all.

The Google style is "TODO(jerry):", where the name says who to ask for more information, not necessarily who'll fix it or who wrote the comment. In this case, the task is simply to internationalize the strings by putting them into resources.

phil-lopreiato: I mean, what does `I18N` mean?

1fish2: `I18N` is short for "internationalization" ("i" followed by 18 letters followed by "n"). It's goofy but common practice, along with `L10N` for "localization".

android/src/main/java/com/thebluealliance/androidclient/helpers/MatchHelper.java
1fish2: Agreed.

How do you make static methods get a Context needed to look up resources?

[One trick](http://stackoverflow.com/a/5114361/1682419) is to have the Application instance save it in a static variable, but is that reliable for use within a BroadcastReceiver?

android/src/main/res/values/strings.xml
1fish2: I'll fix that momentarily.

android/src/main/java/com/thebluealliance/androidclient/helpers/MatchHelper.java
phil-lopreiato: You should just be able to make it a method parameter.

BroadcastReceiver's [onReceive](https://developer.android.com/reference/android/content/BroadcastReceiver.html#onReceive%28android.content.Context,%20android.content.Intent%29) has the current context as a parameter, so it'll be available.

android/src/main/java/com/thebluealliance/androidclient/helpers/MatchHelper.java
phil-lopreiato: Ah, got it

test_notification.py
1fish2: Thanks. It turned out well after solving the permissions puzzle and the quoting puzzle through 2 shells + adb. Two Stack Overflow links in the comments.

android/src/main/java/com/thebluealliance/androidclient/helpers/MatchHelper.java
1fish2: Will do.

upcoming_match.py
1fish2: This file has some small utility as an example or if you don't have/get scriptine. I could go either way on whether to keep it.

android/src/main/AndroidManifest.xml
1fish2: fixed

android/src/debug/res/values/strings.xml
1fish2: fixed

android/src/main/java/com/thebluealliance/androidclient/datafeed/deserializers/EventDeserializer.java
1fish2: Done. This makes the code more DRY and more resilient, but one could go further to check that more JSON fields are present and contain a JsonArray or whatever type is needed.

1fish2:
I updated the PR with all the fixes except the Fragment#setArguments() thing. I'll get to that later today.

Meanwhile, does this project use the deprecated https://github.com/evant/android-studio-unit-test-plugin or its replacement http://tools.android.com/tech-docs/unit-testing-support ?

phil-lopreiato:
It seems like newer versions of android studio have finally enabled running unit treats on the local JVM. I'll take a look at it, and we'll decide what to do. Either way, it should be a separate PR

1fish2:
Yes. For now, should I install the deprecated AS plug-in and library and figure out how to use them?

phil-lopreiato:
Yeah, that'll be the best for now. All the tests have their own package, and they're passing (travis runs them automatically), so I don't think there's anything you'll have to do.

android/src/main/java/com/thebluealliance/androidclient/fragments/tasks/UpdateUserModelSettingsTaskFragment.java
1fish2: setArguments() requires making ModelNotificationFavoriteSettings Parcelable so it can serialize into a Bundle, right? Will that work? I.e. UpdateUserModelSettings won't try to return values by storing them into that instance?

1fish2:
The Fragment fix is in and that should handle the rest of the feedback so far.

I'm not sure how to test the Fragment fix, though. I set a breakpoint on the no-arg UpdateUserModelSettingsTaskFragment() constructor and tried rotating various setting screens but didn't trigger the breakpoint.

phil-lopreiato:
Oh, if the fragment is created via constructor from within another, the system won't create it automatically, so it can use the original one, like you're seeing.

1fish2:
I don't fully understand that. When the framework recreates the Fragment, it always calls the no-arg constructor, which should hit a breakpoint there as long as I'm in the right screen when rotating the display. Which screen is that?

Also, how soon can the Fragment call getArguments() successfully? Is the constructor too soon? The Android docs don't say. The web has a lot of talk about getArguments() but not much clarity. Many examples call it from onCreate(), so I guess I'll change it to do that.

phil-lopreiato:
You can call getArguments in onCreate (it won't work in the constructor, because you don't call setArgs until after the fragment has been created).

Some of those 0-arg constructors aren't being called because the fragments aren't constructed by the Android framework. For example, UpdateUserModelSettings is constructed in an onClick handler, so it won't be affected by device rotation. That's a case where we can keep the multi-arg constructor, but have a 0 argument one so the compiler doesn't complain about it not being there.

1fish2:
`onAttach()` gets called before `onCreate()`. Is that too soon to call `getArguments()`?

Anyway *if* this fragment will always go away before the framework might recreate it, then the Bundle code is overkill.

phil-lopreiato:
> onAttach() gets called before onCreate(). Is that too soon to call getArguments()?
I think onAttach is okay, as well. 

> Anyway if this fragment will always go away before the framework might recreate it, then the Bundle code is overkill.
That's true. A quick 'find usages' should answer that (on my phone, can't check now)

1fish2:
Find Usages indicated that the fragment gets created by (1) FABNotificationSettingsActivity#onClick() [what's "FAB"?] with a fabHandler.postDelayed Runnable to remove it, and (2) MyTBAModelSettingsActivity#onClick(), without code to remove it.

And I can't say if there's a chance of future code creating and keeping this fragment.

1fish2:
I merged the latest changes from master.

What remains to do to merge this PR?

phil-lopreiato:
It looks good to me. I'll test it out tomorrow morning and merge it.

Test failures don't look to be your fault. It should be merged into the `2015-comp-freeze` branch instead of `master`, but I can do that at merge time.

1fish2:
SGTM.
I'm about to send the following PR that fixes #382 and #384 . Should I merge `2015-comp-freeze` into that branch first? Or send the PR related to that branch?

phil-lopreiato:
Send it against 2015-comp-freeze and make sure it'll merge without errors. Don't have it based off master, because some new stuff is starting to get merged there

1fish2:
uh oh. I already merged from master. Should I back that out and merge from 2015-comp-freeze instead?

phil-lopreiato:
Don't worry about it. I'll revert any commits that don't need to be there

1fish2:
Thanks!

phil-lopreiato:
Merged in 3086b85f661fd9eca61d0a1123afb0304950fadb

gregmarra:
🎉

PR #390, opened 2015-04-14
Fix #380 Short event names in notifications

### LINKED ISSUE #380: Short event names in notifications
Although the Event activity shows short event names like "Silicon Valley", "Philomath", "Pacific Northwest", etc, notifications show long names like "Silicon Valley Regional sponsored by Google.org", "PNW District - Philomath Event", and "Pacific Northwest District Championship".

This is not great for the small display area available and it makes for excess text in frequent match_score and upcoming_match notifications.

The server could send the short event names along with the long event names (as it does for alliance_selection notifications) or instead of the long event names. Or the client could query its local DB for the short names. (Would asynchronous queries be a problem for notification handling?)

1fish2:
This computes short event names on the client using string patterns similar to the server's code.

We could move the strings into resources, but it's not clear if that's enough to internationalize the code and it's awkward for a utility class to load string resources. Is it even possible at class loading time?

After way too much futzing I punted on Android Studio's test runner and just used Gradle:

* AS doesn't think the test classes are in the code base, unless one renames android/src/test/ to android/src/androidTest/
* AS doesn't think junit.jar is in the module.
* Trying the [new testing instructions](http://tools.android.com/tech-docs/unit-testing-support), the first step is to set the gradle version to 1.1.0. Even with that, AS doesn't show the Experimental > Enable Unit Testing Support checkbox. Nor does it let one set Test Artifact to Unit Tests, in fact hovering over that pop-up menu says you have to update to gradle 1.1.0 to use it! And gradle sync complains "No such property: bootClasspath for class: com.android.build.gradle.LibraryPlugin".
* The Android Studio Unit Test Plugin croaked two or three times so I disabled it.


phil-lopreiato:
I'm surprised you're having trouble with the tests. I check out your branch directly and can run the tests though AS. Travis works too. That's pretty strange...

OTOH, the PR looks good. I'll test it out this afternoon and merge it.

1fish2:
(Sorry to tweak the code during review, but that latest commit is a comment-only change.)

I would like to debug the AS test runner problem. The symptoms are mentioned above.

* Hypothesis 1: Installed software. I'm running Android Studio 1.2 beta 3 (the latest in the dev channel), Java 8 "1.8.0_45", up to date Android SDK, on Windows 7.
* Hypothesis 2: Configuration.
    * JUnit Run/Debug configuration "TestDebug in JUnit":  
    Test kind: All in package  
    Package: com.thebluealliance.androidclient.test ["test" is displayed in red and there's a warning about that because AS doesn't think that's a Java package in the project.]  
    Search for tests: In whole project  
    VM options: -ea  
    Working directory: C:\sourcecontrolled\the-blue-alliance-android  
    Before launch: Gradle-aware Make  
    * JUnit Run/Debug configuration "HelpersTest":  
    Test kind: Class  
    Class: com.thebluealliance.androidclient.test.helpers.EventHelperTest ["test" is displayed in red and there's a warning about that because AS doesn't think that's a Java package in the project.]  
    VM options: -ea  
    Working directory: C:\sourcecontrolled\the-blue-alliance-android  
    Use classpath of module: android  
    Before launch: Gradle-aware Make

Gradle tasks like `./gradlew testDebugBlue` work fine. -- No, wait. They were fine yesterday. Gradle just fetched new roboelectric JARs and now it can't find org.robolectric:shadows-core:jar:21:3.0-SNAPSHOT. Ditto for Travis CI. Argh.

phil-lopreiato:
The robolectric thing happens often because we need to run snapshot versions for Lollipop support. It should hopefully be back soon?

Here's my setup:
 
- Android Studio 1.0.0 JDK 1.18.0_25, Fedora 21 kernel 3.19.3
- I'd venture to guess that the newer android studio's testing support has broken something in the past (which is why I've refused to update mine, at least until after champs)

I think there's a way to set the testing directory in `build.gradle` via sourcesets

phil-lopreiato:
Wait, figured it out. Change the robolectric version to `3.0-rc2` and it should build again.

phil-lopreiato:
Perfect, I'll merge this now. We can still discuss your Android Studio test config here, though. I'm willing to bet that the update broke something.

1fish2:
You are right!

Downloading the AS 1.0 installer actually installs 1.1.0. After reinstalling the AS Unit Test Plugin (it crashes in AS 1.2 beta 3), the test runner works!

AS treats the test directory as Java code and notices that EventHelperTest.java no longer needs to import org.junit.Assert. (It's leftover from a test method that just called fail() to test the test runner.)

The new AS built-in unit testing might possibly work after removing the android-unit-test stuff from the gradle files.

phil-lopreiato:
Happy you got it figured out! It seems we'll have a tricky migration once we update (this is an after champs process). I'd like to keep robolectric as I'm betting it's more fully featured, but the android-unit-test plugin may not be needed anymore.

PR #393, opened 2015-04-15
Fix #392 Notification Testing Script Temporarily Disabled

### LINKED ISSUE #392: Notification Testing Script Temporarily Disabled
Had to remove the dynamic permission from the manifest to get the Play Store to accept the build.

@1fish2 is there any other way to accomplish that?

See details in d6abe304a55462f7e1a8ef3cbb90f9fca46d428f

1fish2:
Play Store doesn't like debug/release string resources for enabling GCM
testing in debug builds, so do it instead via Manifest Merger placeholders.

Changing this setting in build.gradle does work although I can't build a
release config to verify there.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -29,6 +29,7 @@ android {
             debug {
                 versionNameSuffix "/"+gitBranch+"/" + gitSha
                 applicationIdSuffix ".development"
+                manifestPlaceholders = [gcmPermissionRequired: ""] // "" => let GCMBroadcastReceiver accept Intents from 'adb shell am broadcast'
             }
 
             // run 'gradlew assembleDebugBlue' to do a debug signed build without using debug resources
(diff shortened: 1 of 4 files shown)


android/build.gradle
```diff
@@ -38,10 +39,12 @@ android {
                 signingConfig signingConfigs.debug
                 versionNameSuffix "/"+gitBranch+"/" + gitSha
                 applicationIdSuffix ".development"
+                manifestPlaceholders = [gcmPermissionRequired: "com.google.android.c2dm.permission.SEND"]
```
phil-lopreiato: This should be the empty string. The debugBlue should be a totally functional debug blue, except it uses prod styling.

android/src/main/AndroidManifest.xml
```diff
@@ -200,7 +200,7 @@
         </receiver>
         <receiver
             android:name="com.thebluealliance.androidclient.gcm.GCMBroadcastReceiver"
-            android:permission="com.google.android.c2dm.permission.SEND">
+            android:permission="${gcmPermissionRequired}">
```
phil-lopreiato: Can you just comment that this is defined in build.gradle so it's easy to find in the future?

phil-lopreiato:
The Play Store didn't complain about uploading this branch, so I think this is the correct approach

1fish2:
Excellent! Thanks for testing it on Play Store.

phil-lopreiato:
Tests passed :+1:

PR #395, opened 2015-04-15
Fix #387 Incorrect team@event Summary Alliance Determination

### LINKED ISSUE #387: Possible Incorrect team@event Summary Alliance Determination
From comments on 9d7256f3ffb0c53174aebe38074e9acf8a3c19a (https://github.com/the-blue-alliance/the-blue-alliance-android/commit/9d7256f3ffb0c53174aebe38074e9acf8a3c19af#commitcomment-10676779)

1fish2:
Refactor out Match hasTeam() to apply the fix everywhere.

Refactor out red/blue alliance/teams/score getters. (I almost made them instance methods but adjusting the exception-handling code would not be as local.)

Use the existing Match#hasBeenPlayed() method.

### DIFF
--- a/android/src/test/java/com/thebluealliance/androidclient/test/models/MatchTest.java
+++ b/android/src/test/java/com/thebluealliance/androidclient/test/models/MatchTest.java
@@ -6,6 +6,7 @@ import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
 import com.thebluealliance.androidclient.models.Match;
 
+import org.junit.Before;
 import org.junit.runner.RunWith;
 import org.robolectric.RobolectricTestRunner;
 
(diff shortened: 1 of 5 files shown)


android/src/test/java/com/thebluealliance/androidclient/test/models/MatchTest.java
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Created by Jerry on 4/15/2015.
+ */
+@RunWith(RobolectricTestRunner.class)
+public class MatchTest {
+    private static final String ALLIANCES_2009 = "{" +
+            "\"red\":  {\"score\": 81, \"teams\": [\"frc68\", \"frc217\", \"frc247\"]}," +
+            "\"blue\": {\"score\": 98, \"teams\": [\"frc111\", \"frc971\", \"frc67\"]}" +
+            "}";
+    private static final JsonObject ALLIANCES = new JsonParser().parse(ALLIANCES_2009).getAsJsonObject();
```
phil-lopreiato: You can do this in a [@Before](http://junit.sourceforge.net/javadoc/org/junit/Before.html) annotated method to make sure done right before each test has been run.

phil-lopreiato:
Looks good! It doesn't really matter if you move the tests to a before annotated method (I personally think it's cleaner, but it'll work either way) - let me know which you'd like to do and I'll merge when you're ready

android/src/test/java/com/thebluealliance/androidclient/test/models/MatchTest.java
1fish2: Since JsonObject is mutable, it'd be better to create it in a @Before method. I'll do that now.

1fish2:
Ready.

phil-lopreiato:
:+1:

PR #410, opened 2015-04-24
Fix #406 show the day-of-week with each match time

### LINKED ISSUE #406: match schedule view doesn't mention the day
In a CMP match schedule like this:
match-schedule
you have to look closely to notice that the match times jump backwards, no doubt across a day boundary.

It'd be nice to put a visual indicator (bold line or color change or something) at the day boundary (an overnight gap in the match schedule) and also show the day of week next to the match time (in the viewer's potentially distant time zone) whenever it differs from the preceding entry.

1fish2:
This distinguishes today's matches from tomorrow's matches and from
yesterday's matches with delayed results (e.g. right now on Carson and
Tesla).

I tried out two approaches. This simple, low-risk approach just prepends
the day-of-week to the formatted time. Would any locales rather append the
day-of-week?

The other approach uses the user's 12/24 hour preference to select the date
format pattern "E HH:mm" or "E hh:mm a" from a string resource, calls
DateFormat.getBestDateTimePattern() on API level >= 18 to allow
locale-specific adjustments and field reorderings, then uses that to format
the Date. Localizing those resources might fight with
getBestDateTimePattern(), so maybe the code should use one of those two.
The app isn't really internationalized yet so the added complexity doesn't
pay off. I stashed this approach and went back to the simple approach.

A match list might look less busy if entries omitted the day-of-week when
following a list entry with the same day-of-week, but then their day/time
strings wouldn't line up vertically. Also it's not obvious how to pass that
info across adjacent entries, and it's not a time for avoiding risk.

### DIFF
--- a/android/src/main/java/com/thebluealliance/androidclient/views/MatchView.java
+++ b/android/src/main/java/com/thebluealliance/androidclient/views/MatchView.java
@@ -1,5 +1,6 @@
 package com.thebluealliance.androidclient.views;
 
+import android.annotation.SuppressLint;
 import android.content.Context;
 import android.graphics.Typeface;
 import android.text.format.DateFormat;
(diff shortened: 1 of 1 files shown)


phil-lopreiato:
I can't think of any locales that wouldn't put day of week first - I think the simple approach is fine.

PR #416, opened 2015-04-25
Fix #412 Summary notification shows "@17041057" after 7 items

### LINKED ISSUE #412: Summary notification shows "@17041057" after #7
When a summary notification exceeds 7 entries, it shows `@17041057` for the rest.

I verified that the app is passing the right data to `NotificationCompatBuilder`, and indeed a web search shows that [it's an Android 5.1 bug](https://code.google.com/p/android/issues/detail?id=168890).

18-notifications- 17041057

The bug does not appear to be in `NotificationCompatJellybean` so I think the choices are

1. Live with it and hope Google fixes it.
1. Rather than adding 8+ items to a summary notification on Android 5.1, add 6 items plus a "more..." line. That seems reasonable, although not a big deal.

1fish2:
Limit the summary list to 7 items to work around an Android 5.1 bug and
generally avoid taking up the whole notification shade.

(No hurry to push this to production.)

### DIFF
--- a/android/src/main/java/com/thebluealliance/androidclient/gcm/notifications/SummaryNotification.java
+++ b/android/src/main/java/com/thebluealliance/androidclient/gcm/notifications/SummaryNotification.java
@@ -20,6 +20,11 @@ import java.util.ArrayList;
  * Created by phil on 2/5/15.
  */
 public class SummaryNotification extends BaseNotification {
+    /**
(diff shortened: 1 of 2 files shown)


phil-lopreiato:
LGTM. I don't think we'll push another update this weekend. The next release will probably be right before offseasons start

PR #421, opened 2015-05-24
Updated FAB animations

nwalters512:
This brings the FAB show/hide animation that was used in the Gameday activity to the other activity with FABs. Also fixes some bugs with showing/hhiding the fab that were caused by swiping rapidly between tabs.


phil-lopreiato:
LGTM :+!:

PR #422, opened 2015-05-25
Elevate all the things!

nwalters512:
This adds elevation to all toolbars and tabs in the app to be more consistent with the material guidelines. As this uses the elevation APIs introduced with Lollipop, the elevation is only visible on 5.x+ devices. Also changes how the status bar color is set in activities with myTBA settings panels so that the status bar retains its transparency.


phil-lopreiato:
Merge conflicts, but LGTM

phil-lopreiato:
Travis error after merge: https://travis-ci.org/the-blue-alliance/the-blue-alliance-android/builds/64166411#L1684

Are all the resources added?

nwalters512:
Yep, I just triple checked. Ughhhh.

phil-lopreiato:
That's not even an error I've seen before :/

nwalters512:
Huh... it passed!

phil-lopreiato:
¯\_(ツ)_/¯

PR #429, opened 2015-06-02
More fixes, cleanup, and updates

nwalters512:
* Bumps all dependencies to their latests versions
* Reformatted all code and resources
* Fixed several small bugs
* Replaced deprecated classes with their updated counterparts


phil-lopreiato:
:+1:

PR #431, opened 2015-06-04
Lots of Database Updates

### LINKED ISSUE #398: [App v3] Database Updates
We should improve the design of the db this offseason

Ideas:
 - ~~Remove search index tables. Make a row for `indexes` in Event/Team tables and query that using projection instead of having to worry about multiple tables~~
 - Add foreign key constraints to the other tables (eventTeam, districtTeam, media, match, award) to relate to instances of `teamKey` and `eventKey`

phil-lopreiato:
**DO NOT MERGE YET**

Requires a min API version bump to 16. This should only affect 2% of installs, so we're okay here

Closes #398 

TODO:
 - Add foreign key constraints where we can
 - ~~javadoc~~
 - Test as much of the db as possible (queries, cursor inflation, etc)

### DIFF
--- a/android/src/main/java/com/thebluealliance/androidclient/datafeed/Database.java
+++ b/android/src/main/java/com/thebluealliance/androidclient/datafeed/Database.java
@@ -558,6 +558,10 @@ public class Database extends SQLiteOpenHelper {
             mDb.rawQuery("DELETE FROM " + getTableName(), new String[]{});
         }
 
+        public void deleteSearchIndex(Team team){
+            deleteCallback(team);
+        }
+
         public void recreateAllSearchIndexes(List<Team> teams){
             mDb.beginTransaction();
             try{
@@ -651,6 +655,10 @@ public class Database extends SQLiteOpenHelper {
 
         public void deleteAllSearchIndexes(){
             mDb.rawQuery("DELETE FROM " + getTableName(), new String[]{});
+        }
+
+        public void deleteSearchIndex(Event event){
+            deleteCallback(event);
         }
 
         public void recreateAllSearchIndexes(List<Event> events){
--- a/android/src/main/java/com/thebluealliance/androidclient/datafeed/Database.java
+++ b/android/src/main/java/com/thebluealliance/androidclient/datafeed/Database.java
@@ -33,8 +33,6 @@ import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.Semaphore;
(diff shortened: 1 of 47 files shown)


android/src/main/java/com/thebluealliance/androidclient/datafeed/Database.java
```diff
@@ -190,99 +188,82 @@
...
+    private Teams mTeamsTable;
+    private Events mEventsTable;
+    private Awards mAwardsTable;
+    private Matches mMatchesTable;
+    private Medias mMediasTable;
+    private EventTeams mEventTeamsTable;
+    private Response mResponseTable;
+    private Districts mDistrictsTable;
+    private DistrictTeams mDistrictTeamsTable;
     private Favorites favoritesTable;
```
nwalters512: For consistency's sake, should probably be ```mFavoritesTable```.

android/src/main/java/com/thebluealliance/androidclient/datafeed/Database.java
```diff
@@ -435,11 +416,11 @@ public long storeResponse(String url, String updated) {
             cv.put(Response.URL, url);
             cv.put(Response.LASTUPDATE, updated);
             cv.put(Response.LASTHIT, new Date().getTime());
-            return safeInsert(TABLE_API, null, cv);
+            return mDb.insert(TABLE_API, null, cv);
         }
 
         public APIResponse<String> getResponse(String url) {
-            Cursor cursor = safeQuery(TABLE_API, new String[]{Response.URL, Response.LASTUPDATE, Response.LASTHIT},
+            Cursor cursor = mDb.query(TABLE_API, new String[]{Response.URL, Response.LASTUPDATE, Response.LASTHIT},
```
nwalters512: Perhaps refactor to ```getResponseIfExists()``` and remove ```responseExists()```. The only time I see us using ```responseExists()``` is to check if a response exists before trying to retrieve it; combining all of that into one step would save precious CPU cycles.

android/src/main/java/com/thebluealliance/androidclient/interfaces/ModelTable.java
```diff
@@ -1,20 +1,143 @@
...
+
+    private SQLiteDatabase mDb;
+
+    public ModelTable(SQLiteDatabase db){
+        mDb = db;
+    }
+
+    public long add(T in) {
+        mDb.beginTransaction();
+        long ret;
```
nwalters512: document what `ret` is, it isn't immediately obvious from reading through the code.

PR #433, opened 2015-06-05
Added new framework for refreshing

nwalters512:
Ignore the branch name, I had much greater ambitions for this at first.

This breaks the functionality provided by `RefreshableHostActivity` into an object called `RefreshManager`. It keeps much of the same functionality, with the benefit that the logic lives in its own class that can be dropped into anything; as phil says, "Composition over Inheritance". It proved too much of a hassle to integrate into existing code when we would just have to rewrite it all later, so I'm just committing the important parts here and we can wire everything up in the rewrite.

Behavior changes from `RefreshableHostActivity`:
- Will not start a refresh automatically when the activity is resumed; refreshable things should be responsible for their first load of data. `RefreshManager` should be used primarily to deliver notification requests from the user when they click on a refresh button.
- Refreshable things must notify the RefreshManager whenever their refreshing state changes.
- `onRefreshStop` was removed because the user isn't actually able to stop a refresh in progress. Refreshable things should decide for themselves when to stop or cancel any loading tasks.


phil-lopreiato:
:+1:

PR #445, opened 2015-08-11
Design support library

nwalters512:
Switched to using the FAB from the Design Support Library. Also began the transition to using Snackbars instead of Toasts; so far this is limited to the success/fail/noop result of updating mytba model settings.


phil-lopreiato:
:+1:

PR #434, opened 2015-06-09
Revamped info pages

nwalters512:
!!!DO NOT MERGE!!!

Revamps the team/event info pages to utilize a list of cards similar to the one found in the stock contacts app. So far I've only done this for teams, but I'll be doing events next. I'll keep this PR updated with my progress.

Screenshots:

Imgur

Imgur

PR #446, opened 2015-08-12
Updated info views

nwalters512:
Same as #434, but this one targets appv3-datafeed instead, now that this branch has been updated to work with the new datafeed.


phil-lopreiato:
Can you just recolor the social icons to add some differentiation? I think we can part with the bar for it

nwalters512:
Should be good to go.

PR #448, opened 2015-08-12
More no data views

nwalters512:
Added some more "no data" messages and fixed an issue with the myTBA fav/sub lists


phil-lopreiato:
:+1: post travis

PR #451, opened 2015-08-13
Use Retorift 2.0 and Be Smarter

phil-lopreiato:
- Uses Retrofit v2.0 (which is still in `SNAPSHOT` mode). Hopefully it's stable enough...
 - Make Retrofit API return `Response<Observable<blah>>` so we can get caching info out of that
 - Only call the database writer if the caching info indicates the response is new

(still some more cleanup to do, but in a decent place now)


phil-lopreiato:
Shoot, wrong target branch

Issue #453, opened 2015-08-14
.SQLiteException: no such table: tempevents


nwalters512:
Reported via Google Play

App version: 2.2.4

java.lang.RuntimeException: An error occured while executing doInBackground()
at android.support.v4.content.ModernAsyncTask$3.done(ModernAsyncTask.java:137)
at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:355)
at java.util.concurrent.FutureTask.setException(FutureTask.java:222)
at java.util.concurrent.FutureTask.run(FutureTask.java:242)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
at java.lang.Thread.run(Thread.java:818)
Caused by: android.database.sqlite.SQLiteException: no such table: tempevents (code 1): , while compiling: SELECT events.rowid as '_id',key,name,shortName,eventType,eventDistrict,startDate,endDate,location,venue,official,districtString FROM events JOIN tempevents ON tempevents.tempkey = events.key ORDER BY ? DESC
at android.database.sqlite.SQLiteConnection.nativePrepareStatement(Native Method)
at android.database.sqlite.SQLiteConnection.acquirePreparedStatement(SQLiteConnection.java:1093)
at android.database.sqlite.SQLiteConnection.prepare(SQLiteConnection.java:670)
at android.database.sqlite.SQLiteSession.prepare(SQLiteSession.java:588)
at android.database.sqlite.SQLiteProgram.<init>(SQLiteProgram.java:59)
at android.database.sqlite.SQLiteQuery.<init>(SQLiteQuery.java:37)
at android.database.sqlite.SQLiteDirectCursorDriver.query(SQLiteDirectCursorDriver.java:44)
at android.database.sqlite.SQLiteDatabase.rawQueryWithFactory(SQLiteDatabase.java:1437)
at android.database.sqlite.SQLiteDatabase.rawQuery(SQLiteDatabase.java:1376)
at com.thebluealliance.androidclient.activities.MoreSearchResultsActivity$3.loadInBackground(MoreSearchResultsActivity.java:149)
at com.thebluealliance.androidclient.activities.MoreSearchResultsActivity$3.loadInBackground(MoreSearchResultsActivity.java:146)
[...]

PR #456, opened 2015-08-16
Search bugfix

nwalters512:
Looks like we didn't update everything when we changed how search is done. This fixes bug #453.


phil-lopreiato:
Travis.

You should change in `android/build.gradle` the dependency of the jacoco task to be `testDevDebug` instead.

PR #460, opened 2015-09-05
Fixes #459; improves behavior of team year selection

### LINKED ISSUE #459: Invert ordering of team years participated
In appv3-datafeed, the "Select Year" dialog in the teams activity shows years in ascending order. Recent years are probably more relevant to most people, so I think the list should go in reverse chronological order.

nwalters512:
Fixes #459 so that years are now ordered in reverse chronological order.

I also moved around some of the stuff that deals with year selection. This slightly improves separations of concerns.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -187,8 +187,8 @@ dependencies {
 
     // Square Libraries
     compile 'com.squareup.picasso:picasso:2.5.2'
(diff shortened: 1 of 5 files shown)


phil-lopreiato:
:+1:, but fix tests to reflect the change

nwalters512:
Good catch, will do.

PR #461, opened 2015-09-05
Fixed issue where events lists weren't being bound

nwalters512:
If the events list pager was set to position 0 (the default), an onPageSelected event was never fired for the FragmentBinder, and the proper tabs weren't bound. This fixes that issue by manually calling onPageSelected when the proper tab is computed and selected.


phil-lopreiato:
Per travis, one more Retrofit change in `modules/DatafeedModule.java:39`

:+1: otherwise

nwalters512:
I made that change in my other PR; once that's merged, I'll merge those changes into this PR so the tests pass.

phil-lopreiato:
Alright, merged #460, you should just have to rebase and be all set

phil-lopreiato:
@nwalters512 what's up here? Should I merge everything in or do you want to take a look at it again?

PR #474, opened 2015-09-17
FAB layout fixes

nwalters512:
This should fix the weird FAB layout behavior we were seeing on 4.x devices. It turns out that if FABs are placed as a direct child of a CoordinatorLayout, the CL will translate the FAB to account for the extra padding that's added on 4.x devices to allow for the shadow. So, we still have to provide a margin for the FAB.

People with 4.x devices, please test this! I was able to test 2/3 activities with FABs, but not the myTBA settings activity that's launched from the myTBA fav/sub lists (I was running on an emulator where I couldn't enable myTBA).

### DIFF
--- a/android/src/main/res/layout/activity_gameday.xml
+++ b/android/src/main/res/layout/activity_gameday.xml
@@ -32,15 +32,6 @@
                 android:layout_height="match_parent"
                 android:layout_below="@id/tabs" />
 
-            <android.support.design.widget.FloatingActionButton
-                android:id="@+id/filter_button"
-                android:layout_width="wrap_content"
-                android:layout_height="wrap_content"
-                android:layout_alignParentBottom="true"
-                android:layout_alignParentRight="true"
-                android:layout_margin="@dimen/fab_margin"
-                android:src="@drawable/ic_filter_list_white_24dp"
-                app:fabSize="normal" />
         </RelativeLayout>
 
         <LinearLayout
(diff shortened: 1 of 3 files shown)


fangeugene:
FABs are disappearing for me. They show up (with proper margins now!) but then disappear a few seconds later.

fangeugene:
Nevermind, the disappearing FAB issue was because myTBA was not enabled. Everything looks good.

nwalters512:
That being said, we should probably fix this behavior for those of our users who choose not to enable myTBA. The FAB should be set to "invisible" in the layout file so it isn't rendered when the activity is launching. Then, we should show it with a nice scale animation once we're sure the user has myTBA enabled and should therefore see the FAB.

fangeugene:
This is the first time I've seen the FAB show up and then disappear. I've had myTBA disabled before, and this didn't happen. Maybe it's because I'm running it on older hardware this time, assuming nothing in the code changed.

nwalters512:
I noticed the same thing when I was running it in a 4.4 emulator, but I can't reproduce it on 5.1. It might be something different with how Android renders activities while launching? I'm not really sure. Either way, what I suggested should fix that in all cases. I'll try to take care of that tomorrow.

phil-lopreiato:
@nwalters512 can you take this and #461 and put them up again against master?

nwalters512:
I'll work on this a bit tonight. Still need people to investigate stuff on 4.x for me.

phil-lopreiato:
Early next week, I'll flash my Nexus 5 back to KitKat and give you a hand with that

Adam8234:
I have a few tablets running 4.4 KitKat. Do you want me to do testing
instead of flashing your device Phil?

On Sat, Oct 24, 2015 at 1:28 PM, Phil Lopreiato <notifications@github.com>
wrote:

> Early next week, I'll flash my Nexus 5 back to KitKat and give you a hand
> with that
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/the-blue-alliance/the-blue-alliance-android/pull/474#issuecomment-150839667>
> .
>



-- 

Adam C
Programming Expert | FRC Krawler Developer | Team 2052 KnightKrawler

phil-lopreiato:
If you can test, that'll be awesome!

Adam8234:
What branch do you want me to pull?

phil-lopreiato:
@nwalters512?

phil-lopreiato:
Didn't see any issues here on my N5 on 4.4

PR #483, opened 2015-10-24
Onboarding will only show the login prompt if play services is available

nwalters512:
Minor change that shouldn't affect too many people, but it's still a good behavior to have.


phil-lopreiato:
@nwalters512 Can you put this against master? `appv3-datafeed` has already been merged in

PR #485, opened 2015-11-19
Fix myTBA preference, bump platform/libarary versions

nwalters512:
Before, if you clicked to enable myTBA in settings but then canceled it, the preference switch would show that myTBA was enabled until the next time the settings activity was recreated. Now, the PreferenceFragment driving everything will check if myTBA is enabled in onResume and update the UI to reflect that.

This also targets API23 (Marshmallow) and bumps build tools, the support libraries, and Play Services to their latest respective versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ android:
   components:
     - platform-tools
     - tools
-    - build-tools-23.0.1
+    - build-tools-23.0.2
     - android-23
     - extra-android-m2repository
     - extra-google-m2repository
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,15 +6,15 @@ android:
   components:
     - platform-tools
     - tools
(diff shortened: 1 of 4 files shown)


phil-lopreiato:
How much testing have you done for runtime permissions on 23? Because I want to make sure we have the permission auth flow totally down before we up the target (which I've been avoiding :P). Maybe we should do the SDK updates on their own branch for thorough testing. 

The rest seems good to me though

nwalters512:
We don't currently request any permissions in the "dangerous" category (http://developer.android.com/guide/topics/security/permissions.html#normal-dangerous) so the Marshmallow permissions changes don't affect us.

phil-lopreiato:
Oh, perfect, LGTM then.

Can you just update the travis file for SDK 23 so we can check the CI build before I merge?

PR #489, opened 2015-12-02
Better Model Rendering

phil-lopreiato:
[NOT READY YET]

Towards the goal of improving BasicModel#render. Moves rendering logic out of model classes and into dedicated renderers. Next is to delete those paths entirely and do all rendering from a generic renderer and BasicModel.

After that, we can update some ListItems and render all mytba items like they're done elsewhere. Just need to add mytba settings view so event/team/district list items (hide by default)


phil-lopreiato:
Alright, this part is done. Rendering is much more uniform now. 

Next thing is to show mytba data in event/team/district list items, but that's for a separate PR

PR #498, opened 2015-12-05
Team motto

SunnyRaj:


nwalters512:
Can you upload a screenshot of how this looks?

android/src/main/res/values/styles.xml
```diff
@@ -1,7 +1,7 @@
 <resources>
 
     <!-- Base application theme. -->
-    <style name="AppTheme" parent="Theme.AppCompat.Light.DarkActionBar">
+    <style name="AppTheme" parent="Base.Theme.AppCompat.Light.DarkActionBar">
```
nwalters512: I'm curious, why this change? Did the support lib change the namespace?

SunnyRaj: The layout didn't render properly in Android Studio Editor. This was the fix.

phil-lopreiato:
This will depend on https://github.com/the-blue-alliance/the-blue-alliance/pull/1299 in order for data to be shown

SunnyRaj:
Draft I'm working on. Open to comments.

team logo

nwalters512:
Looks solid. I'd like to have an icon for consistency, but I'm not sure if there's a good one to represent the idea of "motto". If not an icon, there should be *something* to indicate that this is the motto, as it might not be immediately clear to an end user.

phil-lopreiato:
I can't think of anything to indicate motto either. I think it'll be alright as is, unless you can think of a good one.

Good stuff @SunnyRaj

nwalters512:
You could try just adding "Motto: " and see how that looks.

1fish2:
Google Image search found this icon for "motto": http://noahsphotography.com/wp-content/uploads/2014/06/MyMottoIcon2.png

phil-lopreiato:
https://github.com/the-blue-alliance/the-blue-alliance/pull/1299 is merged on the server side, I'm going to merge this for now. That icon doesn't look to material-y and I did some polling of people near me and they all guessed that the field meant team motto. So I think it's good.

PR #499, opened 2015-12-06
Butter knife integration

SunnyRaj:
Starts with TeaminfoBinder and EventInfoBinder

### DIFF
--- a/android/src/main/java/com/thebluealliance/androidclient/binders/EventInfoBinder.java
+++ b/android/src/main/java/com/thebluealliance/androidclient/binders/EventInfoBinder.java
@@ -15,12 +15,15 @@ import com.thebluealliance.androidclient.eventbus.ActionBarTitleEvent;
 import com.thebluealliance.androidclient.eventbus.EventRankingsEvent;
 import com.thebluealliance.androidclient.eventbus.EventStatsEvent;
 import com.thebluealliance.androidclient.eventbus.LiveEventMatchUpdateEvent;
+import com.thebluealliance.androidclient.listeners.EventInfoContainerClickListener;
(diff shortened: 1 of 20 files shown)


phil-lopreiato:
Nice! The other ones that needed to be converted are:
 - DistrictPointsListBinder
 - EventTabBinder
 - ExpandableListViewBinder
 - ListViewBinder
 - MatchListBinder
 - StatsListBinder
 - TeamTabBinder

android/src/main/java/com/thebluealliance/androidclient/binders/EventInfoBinder.java
```diff
@@ -57,15 +77,12 @@ public void setInflater(LayoutInflater inflater) {
     //TODO this needs lots of cleanup. Move click events to their own listeners, no findviewbyid
     @Override
     public void updateData(@Nullable Model data) {
-        if (data == null || view == null) {
-            setDataBound(false);
-            return;
-        }
-
+        ButterKnife.bind(this, mRootView);
```
phil-lopreiato: I think that you should break this call out into a new abstract method in AbstractDataBinder called `bindViews`. Each implementing Binder class overrides that methods and just calls `ButterKinfe.bind(this, mRootView)`. Each class needs to do it because the type passed to ButterKnife needs to be the concrete implementing class, not the superclass, for the annotation processing.

Then, modify BaseAPISubscriber to call `bindViews` before it called `updateData`. This way, we can have binders extend from each other (and call `super`'s `updateData`), but only end up calling `ButterKnife.bind` once

phil-lopreiato:
Looks awesome, nice work!

PR #504, opened 2015-12-12
Fixed issue where items were being incorrectly positioned in the toolbar

nwalters512:


android/src/main/java/com/thebluealliance/androidclient/activities/HomeActivity.java
```diff
@@ -55,7 +55,8 @@
 
     private static final String MAIN_FRAGMENT_TAG = "mainFragment";
 
-    @Inject TBAStatusController mStatusController;
+    @Inject
+    TBAStatusController mStatusController;
```
phil-lopreiato: Keep this on one line, per Android Style guide.

You can import it into Android Studio with the instructions [here](https://github.com/the-blue-alliance/the-blue-alliance-android#code-style)

phil-lopreiato:
:+1: after style fix

android/src/main/java/com/thebluealliance/androidclient/activities/HomeActivity.java
nwalters512: Code -> Reformat Code doesn't seem to be fixing this, isn't that the whole point of defining a code style?

nwalters512: Moreover, Reformat Code actually sticks this back on a different line, even after applying the proper code style.

phil-lopreiato: I just pushed a fix in https://github.com/the-blue-alliance/the-blue-alliance-android/commit/90e5e6642492ae86493a55948dcee8e4634394ab, I realized I forgot to set the property. Copy over the new file from master

PR #505, opened 2015-12-12
Show notifications in home activity

nwalters512:
Instead of launching a separate activity, the recent notifications list is now shown in-place in the home activity, list most other top-level lists.

We should talk about if we want to do the same for game day.


android/src/main/java/com/thebluealliance/androidclient/activities/HomeActivity.java
```diff
@@ -55,7 +56,8 @@
 
     private static final String MAIN_FRAGMENT_TAG = "mainFragment";
 
-    @Inject TBAStatusController mStatusController;
+    @Inject
+    TBAStatusController mStatusController;
```
phil-lopreiato: Style

PR #509, opened 2015-12-17
Removed all default "Created by" headers

nwalters512:
It was about time we killed these off once and for all.


phil-lopreiato:
Awesome!

PR #513, opened 2015-12-20
Changed how warning messages are handled and displayed.

nwalters512:
Completely changes how we handle warning messages (offline, FIRST down, etc).

Before, each activity was responsible for overriding `show/hideMessage()` and displaying that message with its own views.

Now, the views that show messages are a part of `NavigationDrawerActivity`'s layout, and all showing/dismissing of messages are handled by `BaseActivity` using those layouts. All possible messages are defined as an enum in BaseActivity. This is to account for the fact that multiple parts of the code may want to show different messages simultaneously. By having a single source of possible message types, we can choose which of the requested messages to show based on their priority. The priority is currently: offline > FIRST down > event down. If we need to add more messages in the future, this system makes it easy.

Screenshot of how they look:

Screenshot of app


android/src/main/java/com/thebluealliance/androidclient/activities/BaseActivity.java
```diff
@@ -31,6 +38,25 @@
 public abstract class BaseActivity extends NavigationDrawerActivity
         implements NfcAdapter.CreateNdefMessageCallback, GoogleApiClient.ConnectionCallbacks, GoogleApiClient.OnConnectionFailedListener {
 
+    public enum WarningMessageType {
```
phil-lopreiato: Refactor this into its own class

phil-lopreiato: Wait, actually use an [IntDef](http://developer.android.com/reference/android/support/annotation/IntDef.html) for this, in its own class. Use the value for the priority

phil-lopreiato:
LGTM. 

Do you agree that we should render regular admin messages (https://github.com/the-blue-alliance/the-blue-alliance-android/pull/515) in this way as well? Or should we go with a dialog sort of thing?

nwalters512:
I think admin messages are more transient than this kind of message, so that might be better implemented as a dialog, or maybe even integrated with the notification list.

phil-lopreiato:
Makes sense. Do you want to give the dialogs a try? I'd try, but not sure it would come out pretty :P

phil-lopreiato:
Also, this has merge conflicts now

1fish2:
This is very helpful but I'd like to suggest rephrasing `This event is down.` to something like `This event is offline.` or `This event's data feed is down.`

nwalters512:
Fixed the merge conflicts. Thanks for the advice @1fish2, does it look better now?

1fish2:
:+1::+1:

phil-lopreiato:
Merge conflicts again :)
Maybe I should've merged this one first...

nwalters512:
I'll take care of this in a bit haha

nwalters512:
Huh, why isn't the travis status showing?

phil-lopreiato:
No idea... I guess #yoloMerge?

phil-lopreiato:
:+1:

PR #514, opened 2015-12-20
Reduces status rate limiting for users who are not logged in

nwalters512:
Because users who aren't logged in won't receive GCM notifications of status updates, we should poll more frequently.

This also sets a separate timeout for showing the "Please upgrade" dialog (we don't want to spam users with that). I would almost feel comfortable making that last days or even weeks. We want to remind users that a new version is available, but don't want to annoy them by asking too frequently. In that case, the "last shown" time should be a persistent preference that will last across application restarts.

### DIFF
--- a/android/src/main/java/com/thebluealliance/androidclient/datafeed/status/TBAStatusController.java
+++ b/android/src/main/java/com/thebluealliance/androidclient/datafeed/status/TBAStatusController.java
@@ -9,13 +9,21 @@ import android.content.SharedPreferences;
 import android.net.Uri;
 import android.os.Bundle;
 import android.support.annotation.Nullable;
+import android.util.Log;
 
 import com.google.android.gms.analytics.Tracker;
 import com.google.gson.Gson;
+
 import com.thebluealliance.androidclient.BuildConfig;
+import com.thebluealliance.androidclient.Constants;
 import com.thebluealliance.androidclient.R;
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/thebluealliance/androidclient/datafeed/status/TBAStatusController.java
```diff
@@ -32,30 +40,50 @@
...
     private final SharedPreferences mPrefs;
     private final Gson mGson;
     private final Tracker mAnalyticsTracker;
+    private final Context mContext;
 
     private long mLastUpdateTime;
     private long mLastDialogTime;
 
+    private boolean mUserIsLoggedIn;
+    @Inject Context context;
```
phil-lopreiato: Don't make this a member variable if you're only using it in the constructor.

phil-lopreiato: @nwalters512 you can remove this one too. You don't need the annotated variable if you're using constructor injection, see [docs](http://square.github.io/dagger/)

nwalters512: Sorry, that was left over from when I was playing around to try to make it work, I forgot to take it out.

PR #517, opened 2015-12-22
Random cleanup

nwalters512:
I was going through some of the codebase and found some stuff I wanted to clean up. This is a catch-all branch for all that stuff.

DON'T MERGE YET. This is a work in progress. I'll just accumulate stuff here until I'm happy with where things are at.

### Contributors

This fixes the contributors list, although not in the proper new-datafeed-fancy-dependency-injection way (Phil's going to take care of that). I also added the contribution count of each user while I was at it.

Imgur

### Status Bar Icons
Thanks to some [nice new APIs](http://developer.android.com/reference/android/view/View.html#SYSTEM_UI_FLAG_LIGHT_STATUS_BAR) we got in Marshmallow, we can tell the system to make the status bar icons a dark color. This looks good when the status bar is a light color, like the myTBA screens, when it is yellow. This pull request makes use of that new API to do just that.

Before:
Imgur

After:
Imgur


nwalters512:
Pending peer reviews, this should be good to merge. Someone should test this on a API<23 device to make sure I didn't screw up anything with the status bar stuff.

PR #542, opened 2016-02-19
Cloud Endpoints Library Updates

phil-lopreiato:
This PR adds the source for the cloud endpoints client library into the source tree. It comes with a script to update.

It also enables myTBA on all debug builds! Now that I've had a closer look at the source for the client library, we can set the root url, so we can change it to a different appspot project ID. Debug project id is read from `tba.properties`, and this also assumes you have the server set up properly.

[Review on Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/542)


android/src/main/java/com/thebluealliance/androidclient/accounts/AccountHelper.java
```diff
@@ -96,8 +97,17 @@ public static GoogleAccountCredential getSelectedAccountCredential(Context conte
         return credential;
     }
 
-    public static TbaMobile getTbaMobile(GoogleAccountCredential credential) {
+    public static TbaMobile getTbaMobile(Context context, GoogleAccountCredential credential) {
         TbaMobile.Builder tbaMobile = new TbaMobile.Builder(HTTP_TRANSPORT, JSON_FACTORY, credential);
+        if (Utilities.isDebuggable()) {
+            // If a debug build, set Cloud Endpoints to point to a dev TBA instance
+            String appspotId = Utilities.readLocalProperty(context, "appspot.projectId");
+            if (!appspotId.isEmpty()) {
```
nwalters512: If the debug appspot ID is empty, will TbaMobile just point to the default site, or should we provide some sensible default for apiUrl?

phil-lopreiato: It defaults to `tbatv-prod-hrd` (main site) if nothing is there. Which won't work because the keys aren't set up, but it's best to have these fail with auth errors, rather than 404.

This is consistent with how it works now

nwalters512:
Reviewed 23 of 23 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/the-blue-alliance/the-blue-alliance-android/542)*

PR #554, opened 2016-02-20
Subscriber Bugfix: Clear data to bind after check

### LINKED ISSUE #551: Issues Opening 2016 Matches
Appears, then disappears. Seems like issues from matches in db

phil-lopreiato:
If the first time it executes goes successfully and we generate some
data to bind, but the second time has some error (like the server side
erroring), keep the old data. Only clear when we know we have something
to replace it with

Fixes #551

[Review on Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/554)

### DIFF
--- a/android/src/main/java/com/thebluealliance/androidclient/binders/EventInfoBinder.java
+++ b/android/src/main/java/com/thebluealliance/androidclient/binders/EventInfoBinder.java
@@ -80,6 +80,13 @@ public class EventInfoBinder extends AbstractDataBinder<EventInfoBinder.Model> {
 
     @Override
     public void updateData(@Nullable Model data) {
+        if (data == null) {
+            if (!isDataBound()) {
+                bindNoDataView();
+            }
+            return;
+        }
+
         mSocialClickListener.setModelKey(data.eventKey);
         mIsLive = data.isLive;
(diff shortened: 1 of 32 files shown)


nwalters512:
Reviewed 27 of 27 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/the-blue-alliance/the-blue-alliance-android/554)*

nwalters512:
LGTM.

All this duplicate code makes me think it would be nice to build this behavior into the base `Subscriber` class instead of having to duplicate the behavior in every single subclass. But you know the data framework better than me, I'm not sure if that's possible.

phil-lopreiato:
We do have `BaseAPISubscriber`, which I tried first, but there was an issue.

Some Subscribers have a custom model (e.g. a wrapper object for a `Match` and an `Event`, say). We only want to run the subscriber when all components are non-null, which we can't test in the superclass because we don't know the types.

The appropriate way to fix this is to allow Subscribers to override a "isDataValid" method or something where they can do all the tests they need (and default to a null check). 

Hmmm, that actually sounds better...

nwalters512:
Nice work, it looks a lot cleaner this way

PR #559, opened 2016-02-21
Enable generic URLs for ping/broadcast notifications

nwalters512:
If for some reason we want to direct users to a non-TBA URL via a ping or broadcast notification, we can now do that! This simply passes off any URL not in the thebluealliance.com domain to a generic View intent, which will fire off the approriate app/browser.

[Review on Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/559)


phil-lopreiato:
Reviewed 2 of 2 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/the-blue-alliance/the-blue-alliance-android/559)*

PR #566, opened 2016-02-23
Fixes #544 (lists scroll to top

### LINKED ISSUE #544: Bug: lists scroll to top when resuming activity
Steps to reproduce:

1) open an event
2) scroll to "Rankings" tab
3) scroll to bottom of list
4) click on last team
5) once the activity loads, click "Back" (not up, that triggers a different issue)
6) observe how after a second, the list jumps back to the top

This actually applies to all tabs in the activity: if you swipe between several tabs, scroll them all down, then do the above process, they will all be scrolled to the top. There's no reason scroll position shouldn't be persisted.

This appears in several places throughout the app, not only in the "Events" activity.

nwalters512:
As best as I can tell, this fixes #544. The fix ended up relating to how data was bound to lists. We used to create a new adapter every time we got new data; setting that new adapter to the list would wipe any scroll state the list had. Now, we reuse the existing adapter, swap in the new data, and tell the adapter its data has changed. The system handles the rest, and everyone is happy.

Also fixes a related bug where the main teams list would reset to the top of the first tab whenever it was resumed. This is also caused by swapping in a new adapter when none is needed. Now, a new adapter is only created a) on the first data update, and b) if the number of tabs changes.

[Review on Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/566)

### DIFF
--- a/android/src/main/java/com/thebluealliance/androidclient/adapters/ExpandableListViewAdapter.java
+++ b/android/src/main/java/com/thebluealliance/androidclient/adapters/ExpandableListViewAdapter.java
@@ -16,30 +16,27 @@ import java.util.List;
 
 public class ExpandableListViewAdapter extends BaseExpandableListAdapter {
 
-    public final List<ListGroup> groups;
-    public LayoutInflater inflater;
+    public final List<ListGroup> mGroups;
+    public LayoutInflater mInflater;
     public Activity mActivity;
 
     private boolean mIsChildSelectable = true;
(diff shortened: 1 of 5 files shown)


phil-lopreiato:
Reviewed 5 of 5 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/the-blue-alliance/the-blue-alliance-android/566)*

PR #578, opened 2016-02-25
Fix buffering for Firebase tickers

phil-lopreiato:
If the ticker contains fewer elements than the existing buffer size,
then it'll hang forever waiting for enough items to come through and
fill the buffer.

Buffering is requried since we're pushing items into the Observable at
"for loop" speed, so we need to be able to slow down the processing (see
ResumableRxFirebase).

This now puts a time limit on the buffers, see
http://reactivex.io/RxJava/javadoc/rx/Observable.html#buffer%28long,%20java.util.concurrent.TimeUnit,%20int%29

The main downside is the onNext callback is executed every `period` of
time (with empty data). This commit has it at every 5 seconds - also the
max initial loading time. Not sure what the perf implications are.

Can some other people test this out before I merge and see if it affects
perf? It seems like it does to me, but I'm not sure if that's
confirmation bias now that I'm looking for it.

[Review on Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/578)


phil-lopreiato:
Actually, I see what the perf problem is. Every 5 seconds or whatever, we're completely trashing the adapter and rebuilding it with all new values. There's got to be a better way to handle that...

PR #587, opened 2016-02-29
Fix view match up navigation

nwalters512:
Long story short, Android doesn't work as documented: http://stackoverflow.com/questions/14602283/up-navigation-broken-on-jellybean

Rather than relying on the framework to navigate for us, we handle it ourselves to get the behavior we want.

[Review on Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/587)


phil-lopreiato:
> Long story short, Android doesn't work as documented

of course it doesn't....

LGTM :+1:

PR #592, opened 2016-03-01
[It's aliiiiiive!] RecyclerView

nwalters512:
DON'T MERGE YET.

This is my first crack at getting a RecyclerView up and running within the context of our app. Things were made a bit tricky by our data/display layers and the fact that everything had to be very generic, but I found an approach that should work well.

My first commit is really just a proof of concept. It properly renders two types of notifications in the recent notifications view. I still have to clean things up, add a bunch of documentation, finish implementing this for the other notification types, etc. Feel free to add any critiques to the approach I'm taking, it's a fairly large departure from how things were done with ListView before.

We can do the transition from ListView to RecyclerView piecemeal, so it may make sense to merge in this base framework for RecyclerView stuff first, and then use other branches to gradually transition everything else to using it.

[Review on Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/592)


android/src/main/java/com/thebluealliance/androidclient/binders/RecyclerViewBinder.java
```diff
@@ -0,0 +1,127 @@
...
+        if (data == null || mRecyclerView == null) {
+            setDataBound(false);
+            return;
+        }
+        if (data.isEmpty()) {
+            setDataBound(false);
+            return;
+        }
+        if (mMapper == null) {
+            throw new RuntimeException(this.getClass().getName() + " does not have a RecyclerViewBinderMapper set!");
```
phil-lopreiato: Can we handle this more gracefully? Throwing a `RuntimeException` will cause crashes. Can we just complain via `Log.e` and bind the noDataView instead?

android/src/main/java/com/thebluealliance/androidclient/binders/RecyclerViewBinder.java
```diff
@@ -0,0 +1,127 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import butterknife.Bind;
+import butterknife.ButterKnife;
+import io.nlopez.smartadapters.SmartAdapter;
+import io.nlopez.smartadapters.adapters.RecyclerMultiAdapter;
+
+public class RecyclerViewBinder extends AbstractDataBinder<List<Object>> {
```
phil-lopreiato: Why is this `List<Object>` rather than `List<T>` done generically?

android/src/main/java/com/thebluealliance/androidclient/viewmodels/ViewModelRenderer.java
```diff
@@ -0,0 +1,11 @@
+package com.thebluealliance.androidclient.viewmodels;
+
+import android.content.Context;
+import android.support.annotation.Nullable;
+import android.support.annotation.WorkerThread;
+
+public interface ViewModelRenderer<VIEWMODEL, ARGS> {
+
+    @WorkerThread
+    @Nullable VIEWMODEL renderToViewModel(Context context, @Nullable ARGS args);
```
phil-lopreiato: Would we rather enforce a parent class/interface for all the ViewModels? That way, we at least know the base class (rather than just `Object`), which may be useful for handling more generic approaches later? 

That way, the return type of this class just has to be `ParentViewModel`

android/src/main/java/com/thebluealliance/androidclient/binders/RecyclerViewBinder.java
nwalters512: Mu implentstion of RecyclerView will straight up not function without a mapper. This will immediately call our attention to any issues we have before we ship. If we hit this point in production, we did something horribly wrong. That being said, NoDataView and Log.e would probably do the job just as well.

android/src/main/java/com/thebluealliance/androidclient/gcm/notifications/BaseNotification.java
```diff
@@ -25,12 +26,16 @@
...
+import com.thebluealliance.androidclient.viewmodels.ViewModelRenderer;
 
 import java.text.DateFormat;
 import java.util.Calendar;
 import java.util.Date;
 
-public abstract class BaseNotification extends ListElement {
+import io.nlopez.smartadapters.views.BindableLayout;
+
+public abstract class BaseNotification<VIEWMODEL> extends ListElement implements ViewModelRenderer<VIEWMODEL, Void> {
```
phil-lopreiato: Since you're always passing `Void` as the args type and if you do what I suggested below by enforcing only a superclass type for ViewModel, you won't need the generic bit here. 

Just add `public abstract ParentViewModel getViewModel(Context c)` to this class

android/src/main/java/com/thebluealliance/androidclient/fragments/RecentNotificationsFragment.java
```diff
@@ -9,22 +9,31 @@
...
 public class RecentNotificationsFragment
-  extends ListViewFragment<List<StoredNotification>, RecentNotificationsSubscriber> {
+        extends RecyclerViewFragment<List<StoredNotification>, RecentNotificationsSubscriber> {
 
-    @Inject Database mDb;
-    @Inject EventBus mEventBus;
+    @Inject
+    Database mDb;
+    @Inject
+    EventBus mEventBus;
```
phil-lopreiato: Note to self: update style guide so that annotations like this are on one line.

android/src/main/java/com/thebluealliance/androidclient/binders/RecyclerViewBinder.java
nwalters512: We no longer need generics. The binding from objects to views is ultimately done by the mapper provided by subclasses of RecyclerViewFragment, which doesn't use generics at all.

android/src/main/java/com/thebluealliance/androidclient/viewmodels/ViewModelRenderer.java
nwalters512: There's virtually no way to do this kind of thing generically. A ViewModel is really just a nice term for some arbitrary object that will be passed to a view and bound to it later. There's pretty tight coupling between a ViewModel and the view that will ultimately display it later.

android/src/main/java/com/thebluealliance/androidclient/fragments/RecentNotificationsFragment.java
nwalters512: Good call.

android/src/main/java/com/thebluealliance/androidclient/viewmodels/ViewModelRenderer.java
phil-lopreiato: Gotcha. I was suggesting using a parent class not for any real generic reason, but to make it obvious that all the `ViewModel` classes share something (and that we have a place to put common code in the future, should we need it) so we don't need the `List<Object>` type like above (instead use `List<ViewModel>`, which make it much more obvious to the casual reader what is contained in the list, even though it doesn't add any functional value). Just a thought

phil-lopreiato:
I'm a fan! Nice job @nwalters512

android/src/main/java/com/thebluealliance/androidclient/viewmodels/ViewModelRenderer.java
nwalters512: Ah, I see what you mean. I actually briefly had ViewModels extending from a base class, but then took it out once I realized said class would have absolutely no purpose, short of making it nice to read the code, like you said. I guess we may as well; you have a point about having a common place for future code, even if I can't envision what that would be at this point.

android/src/main/java/com/thebluealliance/androidclient/gcm/notifications/BaseNotification.java
nwalters512: The only reason I had for using generics for ViewModelAdapter was for type safety. If for some reason we had to know the exact return type of `renderToViewModel()` for a certain model, we could check that instead of having to use a generic `ParentModelView`, which at this point would be a useless empty class.

phil-lopreiato: Are there any cases where we need to render one model into different kinds of `ViewModel`s? If that's something we need to worry about, the empty parent class may be the best way to go. If not, I don't think it'll make a difference.

nwalters512:
The test suite is going to fail until I finish making ViewModels for the rest of the notification types and update the tests accordingly. Ignore that for now

phil-lopreiato:
Haven't had a chance to test this one, but if the beta looks good in a day or two, I'm fine with merging

phil-lopreiato:
Reviewed 5 of 27 files at r1, 7 of 12 files at r2, 62 of 69 files at r3, 4 of 7 files at r4.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[android/src/main/java/com/thebluealliance/androidclient/binders/RecyclerViewBinder.java, line 53 \[r1\]](https://reviewable.io:443/reviews/the-blue-alliance/the-blue-alliance-android/592#-KBjWMqLPeawWsjt8kqG:-KBjWMqMmXGKldCJxdiA:-94508823) ([raw file](https://github.com/the-blue-alliance/the-blue-alliance-android/blob/c0f7425729b86d1aa02e7e3375c924dfa5295695/android/src/main/java/com/thebluealliance/androidclient/binders/RecyclerViewBinder.java#L53)):*
Can we handle this more gracefully?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/the-blue-alliance/the-blue-alliance-android/592)*

nwalters512:
*[android/src/main/java/com/thebluealliance/androidclient/binders/RecyclerViewBinder.java, line 53 \[r1\]](https://reviewable.io:443/reviews/the-blue-alliance/the-blue-alliance-android/592#-KBjWMqLPeawWsjt8kqG:-KDBcwVIweZNMTHqxhH7:-1960020824) ([raw file](https://github.com/the-blue-alliance/the-blue-alliance-android/blob/c0f7425729b86d1aa02e7e3375c924dfa5295695/android/src/main/java/com/thebluealliance/androidclient/binders/RecyclerViewBinder.java#L53)):*
I think fail-fast makes the most sense here. We want the app to very obviously fail during develoent if the mapper is null.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/the-blue-alliance/the-blue-alliance-android/592)*

phil-lopreiato:
@nwalters512 can you fix merge conflicts?

nwalters512:
@phil-lopreiato any other comments before I merge?

phil-lopreiato:
Nope, go for it. Just make sure you test the notifications list, it was a non-trivial merge, I think

phil-lopreiato:
Reviewed 5 of 7 files at r6.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/the-blue-alliance/the-blue-alliance-android/592)*

phil-lopreiato:
Merge conflicts. And I get the following crash when I open Recent Notifications:
```
03-29 11:50:36.667 6245-6245/com.thebluealliance.androidclient.development E/AndroidRuntime: FATAL EXCEPTION: main
                                                                                             Process: com.thebluealliance.androidclient.development, PID: 6245
                                                                                             org.greenrobot.eventbus.EventBusException: Subscriber class com.thebluealliance.androidclient.binders.RecentNotificationsListBinder and its super classes have no public methods with the @Subscribe annotation
                                                                                                 at org.greenrobot.eventbus.SubscriberMethodFinder.findSubscriberMethods(SubscriberMethodFinder.java:67)
                                                                                                 at org.greenrobot.eventbus.EventBus.register(EventBus.java:136)
                                                                                                 at com.thebluealliance.androidclient.fragments.DatafeedFragment.onResume(DatafeedFragment.java:84)
                                                                                                 at android.support.v4.app.Fragment.performResume(Fragment.java:2020)
                                                                                                 at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1107)
                                                                                                 at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1252)
                                                                                                 at android.support.v4.app.BackStackRecord.run(BackStackRecord.java:738)
                                                                                                 at android.support.v4.app.FragmentManagerImpl.execPendingActions(FragmentManager.java:1617)
                                                                                                 at android.support.v4.app.FragmentManagerImpl$1.run(FragmentManager.java:517)
                                                                                                 at android.os.Handler.handleCallback(Handler.java:739)
                                                                                                 at android.os.Handler.dispatchMessage(Handler.java:95)
                                                                                                 at android.os.Looper.loop(Looper.java:148)
                                                                                                 at android.app.ActivityThread.main(ActivityThread.java:5422)
                                                                                                 at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
                                                                                                 at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
```

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/the-blue-alliance/the-blue-alliance-android/592#-:-KE23cIhqjdRxw7pgJOL:-1148142025)*

nwalters512:
I'll take care of that later today.

phil-lopreiato:
Reviewed 17 of 17 files at r7.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/the-blue-alliance/the-blue-alliance-android/592)*

PR #593, opened 2016-03-01
Fix live event next/last match bugs

### LINKED ISSUE #585: Fix Event Next/Last Match
Shouldn't have this case...
screenshot_20160227-180856

phil-lopreiato:
If the eventbus event doesn't give us anything, hide the container for
the views. This fixes the bug in the case where you firt load next/last
match while there are still matches to go, but then resume the activity
after the last match has been played. This will cause the last match
card to be "stale", still showing the old last match. Now we'll hide it
in that case to produce the proper output.

Additionally, check if ButterKnife is bound during the eventbus
callbacks, since we may be in a resume state where we haven't done so
yet. This will fix crashes trying to access a null view in the onEvent
method, which many execute before `updateData` does or before the
framework initially runs `bindViews`.

Fixes #585

[Review on Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/593)

### DIFF
--- a/android/src/main/java/com/thebluealliance/androidclient/binders/EventInfoBinder.java
+++ b/android/src/main/java/com/thebluealliance/androidclient/binders/EventInfoBinder.java
@@ -36,6 +36,7 @@ public class EventInfoBinder extends AbstractDataBinder<EventInfoBinder.Model> {
     private LayoutInflater mInflater;
     private MatchRenderer mMatchRenderer;
     private boolean mIsLive;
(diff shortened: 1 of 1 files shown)


nwalters512:
Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/the-blue-alliance/the-blue-alliance-android/593)*

PR #603, opened 2016-03-05
Add "View Webcast" button to Event Info

phil-lopreiato:
If there is more than one webcast available, open a chooser dialog to select which stream

device-2016-03-05-131034
device-2016-03-05-131020

[Review on Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/603)


nwalters512:
Do we need a separate card? Might make sense to just make it another item in the top card.

I like the dialog though!

phil-lopreiato:
That's a good idea, I'll give it a try without the card

phil-lopreiato:
Yeah, this is much better

device-2016-03-05-133146

PR #604, opened 2016-03-05
Match Details for 2016

phil-lopreiato:
Adds tabs to ViewMatch activity
device-2016-03-05-181735
device-2016-03-05-181311
device-2016-03-05-181455

[Review on Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/604)


android/src/main/res/values/breakdown2016_styles.xml
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <style name="breakdown_base">
```
nwalters512: This could include `layout_columnSpan`, `layout_columnWeight`, `layout_rowSpan`, `layout_gravity`; it looks like those are consistent across all the `GridView`'s children. You could shave a few hundred lines off the layout file this way.

phil-lopreiato: I tried that, but it didn't like the other namespace (since I'm using the suportlib GridView)...

nwalters512: Try it without the `app:` namespace; that's how I created the toolbar default style.

phil-lopreiato: Right you are, that worked. Thanks!

nwalters512: Much cleaner :)

android/src/main/java/com/thebluealliance/androidclient/adapters/ViewMatchFragmentPagerAdapter.java
```diff
@@ -0,0 +1,49 @@
...
+import com.thebluealliance.androidclient.fragments.match.MatchBreakdownFragment;
+import com.thebluealliance.androidclient.fragments.match.MatchInfoFragment;
+
+import android.support.v4.app.Fragment;
+import android.support.v4.app.FragmentManager;
+import android.support.v4.app.FragmentPagerAdapter;
+
+public class ViewMatchFragmentPagerAdapter extends FragmentPagerAdapter {
+
+    public final String[] TITLES = {"Results", "Breakdown"};
```
nwalters512: I've been meaning to move the viewpager titles to string resources; want to start that here?

I'd switch this array to be an array of string resource ints and pass a Context to the constructor. There, just populate a second array of strings with the resolves strings from the context.

phil-lopreiato: That's a good approach, we can start that here

nwalters512:
Looking good!

Minor visual detail: there's no border after the "total" rows; when I was first scrolling through the breakdown, that lack of border make me unconsciously assume that the darker total rows were column headers or something like that. Maybe stick a bit of padding below the total rows to give the sections a little more visual distinction.

PR #629, opened 2016-03-13
Set custom user agent with gzip for GAE

fangeugene:
Seems like 'gzip' only needs to be present in the user agent. Related to #628.
I put a partial user agent generated using the app's version name in constants, and then append '(gzip)' to the end of it in the request interceptor. Is that appropriate? Is that a decent user agent name?

http://googleappsdeveloper.blogspot.com/2011/12/optimizing-bandwidth-usage-with-gzip.html

---
This change is [Review on Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/629)


nwalters512:
I like the more descriptive user agent. If `getUserAgent()` is only used in that one place, it would probably be fine to just have that method return the string with `(gzip)` already appended.

phil-lopreiato:
All looks fine with me. I'm okay with appending gzip later, in case we want to conditionally disable gzip in the future

PR #636, opened 2016-03-18
Update okhttp to latest 3.2.0

nwalters512:
Should help with a few crashes we've been seeing in the Dev Console. And it's always good to be running the latest and greatest code?

---
This change is [Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/636)

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -236,8 +236,8 @@ dependencies {
 
     // Square Libraries
     compile 'com.squareup.picasso:picasso:2.5.2'
-    compile "com.squareup.retrofit:retrofit:${retrofitVersion}"
-    compile "com.squareup.retrofit:adapter-rxjava:${retrofitVersion}"
+    compile "com.squareup.retrofit2:retrofit:${retrofitVersion}"
+    compile "com.squareup.retrofit2:adapter-rxjava:${retrofitVersion}"
     compile 'com.squareup.okhttp3:okhttp:3.2.0'
 
     compile 'com.google.dagger:dagger:2.0.1'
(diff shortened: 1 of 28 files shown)


phil-lopreiato:
IIRC, it's more complicated than that, because the package name changed. Might want to double check that (and check what version retrofit is depending on). I think this was also the blocker for updating to the newest retrofit

nwalters512:
It looks like Retrofit now depends on `3.2.0`: [pom.xml](https://github.com/square/retrofit/blob/master/retrofit/pom.xml#L18).

Good catch with the package name though. Something is keeping around version `2.5.0` of OkHttp in the project, so the build succeeded just fine, but `3.2.0` isn't actually being used. So yeah, there's more work to be done on this.

nwalters512:
I think that should be it. Most of the changes were simply changing the package in the imports. A few of the APIs changed a bit, but those were easy to update.

The app builds, installs, and runs properly with the new library versions.

phil-lopreiato:
Reviewed 28 of 28 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/the-blue-alliance/the-blue-alliance-android/636)*

phil-lopreiato:
@nwalters512 merge conflicts

phil-lopreiato:
Reviewed 2 of 2 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/the-blue-alliance/the-blue-alliance-android/636)*

PR #645, opened 2016-03-23
Tweak No Data binding logic

### LINKED ISSUE #638: “No alliance data found” after receiving “Alliance selections updated” notification
This happened last week for the CAMA event and again today for the WAELL event.

Steps:

1. Receive and open an “Alliance selections updated” notification. (I did this a while after the notification arrived.)
2. Tap through to see the alliance selections.<br/>
--> It showed “No alliance data found”!
3. Tap MATCHES.<br/>
--> It already had results for several QF matches.
4. Tap back to ALLIANCES.<br/>
--> It still showed “No alliance data found”.
5. Tap the refresh button a lot.<br/>
--> About 1/5 of the time it would flash the alliance selections briefly, then erase it. I could read the first team captain number (3711) before the contents disappeared.
6. Try [www.thebluealliance.com](http://www.thebluealliance.com/event/2016waell) <br/>
--> It did show the alliances.

Currently running App Version 3.3.0, build 3/17/16 11:21 PM, commit 22ce4b7. I did not have a laptop at hand to try debugging tools.

phil-lopreiato:
If the binder receives invalid data, but there is already data bound,
that means the following case is most likely:
 - There was data available in the local cache that we can display
 - The network call returned an error/still is cached with an empty
   value

This can happen when the user receives a push notification which writes
some data locally, but the remote endpoint has not had its cache
cleaered yet, causing the display to flicker between correct data and no
data view.

Fixes #638

---
This change is [Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/645)

### DIFF
--- a/android/src/main/java/com/thebluealliance/androidclient/binders/ExpandableListViewBinder.java
+++ b/android/src/main/java/com/thebluealliance/androidclient/binders/ExpandableListViewBinder.java
@@ -1,16 +1,16 @@
 package com.thebluealliance.androidclient.binders;
 
-import android.support.annotation.Nullable;
-import android.util.Log;
-import android.view.View;
-import android.widget.ProgressBar;
-
 import com.thebluealliance.androidclient.Constants;
 import com.thebluealliance.androidclient.R;
 import com.thebluealliance.androidclient.adapters.ExpandableListViewAdapter;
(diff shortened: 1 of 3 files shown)


1fish2:
Still problematic. See #638

PR #665, opened 2016-04-04
Makes the full team name view expandable

### LINKED ISSUE #233: Make full team name (sponsors list) expandable
Like the "What's New" section for an app in the new Google Play app. It also has a neat expansion animation that stretches the paper material. #materialyolo

nwalters512:
For teams with really long full team names (254, for instance), they are now limited to 3 lines at first, and ellipsized. Clicking the full name expands the view to show the full team name.

Fixes #233

---
This change is [Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/665)

### DIFF
--- a/android/src/main/java/com/thebluealliance/androidclient/binders/TeamInfoBinder.java
+++ b/android/src/main/java/com/thebluealliance/androidclient/binders/TeamInfoBinder.java
@@ -129,7 +129,6 @@ public class TeamInfoBinder extends AbstractDataBinder<TeamInfoBinder.Model> {
             teamFullName.setEllipsize(TextUtils.TruncateAt.END);
 
             teamFullNameContainer.setOnClickListener((view) -> {
-                Log.d(Constants.LOG_TAG, "team full name clicked!");
                 toggleFullTeamNameExpanded();
             });
         }
--- a/android/src/main/java/com/thebluealliance/androidclient/binders/TeamInfoBinder.java
+++ b/android/src/main/java/com/thebluealliance/androidclient/binders/TeamInfoBinder.java
@@ -1,12 +1,17 @@
 package com.thebluealliance.androidclient.binders;
 
+import android.animation.Animator;
+import android.animation.AnimatorListenerAdapter;
+import android.animation.ObjectAnimator;
 import android.net.Uri;
 import android.support.annotation.Nullable;
 import android.text.Spannable;
 import android.text.SpannableString;
+import android.text.TextUtils;
 import android.text.style.TextAppearanceSpan;
 import android.util.Log;
 import android.view.View;
+import android.view.ViewGroup;
 import android.widget.TextView;
 
 import com.thebluealliance.androidclient.Constants;
@@ -19,6 +24,8 @@ import butterknife.Bind;
 import butterknife.ButterKnife;
 
 public class TeamInfoBinder extends AbstractDataBinder<TeamInfoBinder.Model> {
+
+    private static final int TEAM_FULL_NAME_COLLAPSED_MAX_LINES = 3;
 
     @Inject SocialClickListener mSocialClickListener;
 
@@ -118,6 +125,13 @@ public class TeamInfoBinder extends AbstractDataBinder<TeamInfoBinder.Model> {
             string.setSpan(new TextAppearanceSpan(mActivity,
                     R.style.InfoItemLabelStyle), 0, 3, Spannable.SPAN_INCLUSIVE_INCLUSIVE);
             teamFullName.setText(string);
+            teamFullName.setMaxLines(TEAM_FULL_NAME_COLLAPSED_MAX_LINES);
+            teamFullName.setEllipsize(TextUtils.TruncateAt.END);
+
+            teamFullNameContainer.setOnClickListener((view) -> {
+                Log.d(Constants.LOG_TAG, "team full name clicked!");
+                toggleFullTeamNameExpanded();
+            });
         }
 
         teamNextMatchLabel.setVisibility(View.GONE);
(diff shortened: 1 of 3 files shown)


phil-lopreiato:
Screenshot plz?

nwalters512:
Better yet: [demo video](https://www.youtube.com/watch?v=Ky8J2KpygKc)!

phil-lopreiato:
:+1:

fangeugene:
Slick! Is there any way we can indicate that it's expandable?

nwalters512:
It's currently done with the ellipsis, which may be kind of hard to see/not super obvious. Looking back at the original issue, I suggested we fade out the bottom line or two. I could look into that still?

phil-lopreiato:
Give it a shot? It [doesn't look too complex](http://stackoverflow.com/questions/6561701/how-to-fade-edge-of-textview-in-android)

nwalters512:
Except that was deprecated in API 14 :P still, should be easy enough with a drawable clinging to the bottom of the view. The problem is actually computing when to show that; TextView doesn't by default offer a way to tell when its content has been ellipsized, and the ways of finding that out are all hacky and don't handle edge cases well. I'll see if I can figure something out though.

fangeugene:
Are fades Material spec? I wouldn't think so.

fangeugene:
Maybe the ellipses are enough.

nwalters512:
The Play Store uses a fade on the What's New section to indicate that there's more. Personally I'm a fan of sticking with the ellipsis, it's simple and anyone that's that interested in a team's full name will probably notice it.

phil-lopreiato:
I'm fine with sticking with an ellipsis

android/src/main/java/com/thebluealliance/androidclient/binders/TeamInfoBinder.java
```diff
@@ -118,6 +125,13 @@ public void updateData(@Nullable TeamInfoBinder.Model data) {
             string.setSpan(new TextAppearanceSpan(mActivity,
                     R.style.InfoItemLabelStyle), 0, 3, Spannable.SPAN_INCLUSIVE_INCLUSIVE);
             teamFullName.setText(string);
+            teamFullName.setMaxLines(TEAM_FULL_NAME_COLLAPSED_MAX_LINES);
+            teamFullName.setEllipsize(TextUtils.TruncateAt.END);
```
1fish2: Perhaps the ideal would ellipsize out the middle to leave the `& school name` tail:

> _aka_ NASA Ames Research Center /
Google / AMEX Plating / Innovation
First ... & Bellarmine College Preparatory

What do you think about using `TextUtils.TruncateAt.START` as an easy approximation?

android/src/main/java/com/thebluealliance/androidclient/binders/TeamInfoBinder.java
```diff
@@ -165,6 +178,55 @@ private void bindNoDataView() {
...
+            ObjectAnimator animation = ObjectAnimator.ofInt(teamFullName, "height", height, newHeight);
+            animation.setDuration(500);
+            animation.addListener(new AnimatorListenerAdapter() {
+                @Override public void onAnimationEnd(Animator animation) {
+                    teamFullName.setMaxLines(Integer.MAX_VALUE);
+                    teamFullName.setMinHeight(0);
+                }
+            });
+            animation.start();
+        } else {
```
1fish2: Share code between the _then_ and _else_ clauses?

nwalters512: Yeah, in retrospect I could have reused code a bit here. The two clauses were originally very different, didn't realize how much they converged as I iterated :P

android/src/main/java/com/thebluealliance/androidclient/binders/TeamInfoBinder.java
nwalters512: I'm not totally comfortable with hiding the ellipsis in the middle of the text. I think people will expect to see it at the end of the content.

1fish2: How does the ellipsis look at the start of the text?

nwalters512: Weird... I can't get any of the other ellipsize modes to work at all on either my 6.0.1 N5 or my N emulator.

1fish2: So many questions about this on SO! setEllipsize() sounds fragile, interacting with other TextView properties like setSingleLine(), setMaxLines(), setLines(), setHorizontallyScrolling(), ... and varying by Android release.

Oh: [setEllipsize()](http://developer.android.com/reference/android/widget/TextView.html#setEllipsize(android.text.TextUtils.TruncateAt)) says

> If setMaxLines(int) has been used to set two or more lines, only END and MARQUEE are supported (other ellipsizing types will not do anything).

If the current code works on older Android releases, just drop my idea to move the ellipses.

http://stackoverflow.com/search?q=setEllipsize+is%3Aquestion+%5Bandroid%5D -- 33 questions on `setEllipsize()`
http://stackoverflow.com/search?q=ellipsize+is%3Aquestion+%5Bandroid%5D -- 429 questions on `ellipsize`

1fish2: ... and at least one known Android bug: [Issue 2254](https://code.google.com/p/android/issues/detail?id=2254).

PR #693, opened 2016-04-13
Fix stats list binder crash with null menu

### LINKED ISSUE #685: Menu NPE
```
java.lang.NullPointerException: Attempt to invoke interface method 'android.view.MenuItem android.view.Menu.findItem(int)' on a null object reference
                                                                                              at com.thebluealliance.androidclient.binders.StatsListBinder.setSelectedList(StatsListBinder.java:78)
                                                                                              at com.thebluealliance.androidclient.binders.StatsListBinder.updateData(StatsListBinder.java:37)
                                                                                              at com.thebluealliance.androidclient.binders.StatsListBinder.updateData(StatsListBinder.java:20)
                                                                                              at com.thebluealliance.androidclient.subscribers.BaseAPISubscriber.lambda$bindData$2(BaseAPISubscriber.java:170)
                                                                                              at com.thebluealliance.androidclient.subscribers.BaseAPISubscriber.access$lambda$2(BaseAPISubscriber.java:0)
                                                                                              at com.thebluealliance.androidclient.subscribers.BaseAPISubscriber$$Lambda$3.call(Unknown Source)
                                                                                              at rx.internal.schedulers.ScheduledAction.run(ScheduledAction.java:55)
                                                                                              at android.os.Handler.handleCallback(Handler.java:739)
                                                                                              at android.os.Handler.dispatchMessage(Handler.java:95)
                                                                                              at android.os.Looper.loop(Looper.java:148)
                                                                                              at android.app.ActivityThread.main(ActivityThread.java:5422)
                                                                                              at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
                                                                                              at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
```

phil-lopreiato:
Fixes #685

---
This change is [Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/693)

### DIFF
--- a/android/src/main/java/com/thebluealliance/androidclient/binders/StatsListBinder.java
+++ b/android/src/main/java/com/thebluealliance/androidclient/binders/StatsListBinder.java
@@ -25,9 +25,15 @@ public class StatsListBinder extends ListViewBinder implements RadioGroup.OnChec
 
     private ListPair<ListItem> mData;
     private Menu mMenu;
+    private @ListPair.ListOption int mSelectedList;
+
+    public StatsListBinder() {
+        mSelectedList = ListPair.LIST0;
+    }
 
     public void setMenu(Menu menu) {
         mMenu = menu;
+        syncSortMenuItemState(mSelectedList);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


nwalters512:
Seems like this would be caused when `setSelectedList()` is called before `setMenu()`. If that's the case, should we make sure that `setMenu()` will also properly sync the menu state with which list is visible?

Maybe break out that functionality into something like `syncMenuItemVisibility()` and call it in both places.

phil-lopreiato:
But `setSelectedList` is only ever called from within `StatsListBinder`. So we unfortunately do that kind of clean solution. So I think the simple null check is as best we can do.

nwalters512:
I don't really see how that would change our ability to do what I described.

phil-lopreiato:
Ah, right you are. I misread your comment the first time...

nwalters512:
LGTM, pending Travis :stuck_out_tongue:

PR #712, opened 2016-04-27
Changed handling of clicks on event items

nwalters512:
Fixes an issue where clicking on an event in a team's event list opened the event view instead of the team@event view. This is my fault for not verifying behavior when I switched the event lists to use RecyclerView.

This better exploits the features of the adapter library we are using; namely, it moved view event (clicks, etc) handling out of the item and into a listener attached to the adapter. See [here](https://github.com/mrmans0n/smart-adapters#assigning-listeners) for reference.

Going forward, as we make the transition to RecyclerView, this is how we should do things. It generally makes list items more reusable. I'd like to write a class that lets us compose listeners so that we can reuse click listeners throughout the app while still having the option of using multiple listeners per list. Something like this:

```java
ViewEventListenerComposer.Builder builder = new ViewEventListenerComposer.Builder();
builder.compose(new EventClickListener(mContext));
builder.compose(new MyTbaDetailsClickListener(mContext));
creator.listener(builder.build());
```

---
This change is [Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/712)

### DIFF
--- a/android/src/main/java/com/thebluealliance/androidclient/binders/RecyclerViewBinder.java
+++ b/android/src/main/java/com/thebluealliance/androidclient/binders/RecyclerViewBinder.java
@@ -8,8 +8,6 @@ import android.widget.ProgressBar;
 
 import com.thebluealliance.androidclient.Constants;
 import com.thebluealliance.androidclient.R;
-import com.thebluealliance.androidclient.adapters.ListViewAdapter;
-import com.thebluealliance.androidclient.listitems.ListItem;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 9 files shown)


android/src/main/java/com/thebluealliance/androidclient/binders/RecyclerViewBinder.java
```diff
@@ -118,11 +116,11 @@ public void unbind(boolean unbindViews) {
...
     protected void createAndInitializeAdapterForData(List<Object> data) {
         SmartAdapter.MultiAdaptersCreator creator = SmartAdapter.items(new ArrayList<>(data));
-        mMapper.initializeMaps(creator);
+        mMapper.initializeAdapterCreator(creator);
         mAdapter = creator.into(mRecyclerView);
     }
 
-    public interface RecyclerViewBinderMapper {
-        void initializeMaps(SmartAdapter.MultiAdaptersCreator creator);
+    public interface RecyclerViewAdapterCreatorInitializer {
```
nwalters512: I hate this name with a passion but idk what else to call it.

phil-lopreiato: Yay for Java!

phil-lopreiato:
Reviewed 9 of 9 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/the-blue-alliance/the-blue-alliance-android/712)*

phil-lopreiato:
I really like using RecyclerView library for click adapters, and the builder bit is awesome.

I'd imaging we define our own interface that returns `true` once the click has been handled, and the built listener just iterates through the list until one of the sub-listeners returns `true`.

nwalters512:
That was exactly what I was thinking!

PR #743, opened 2016-05-21
Use RecyclerView for event rankings

nwalters512:
This change is [Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/743)


nwalters512:
`RankingListElement` is not longer used, but it's still a part of the fragment test (which doesn't work anyways because we're using recyclerview now). Not sure if I should leave that intact or not.

phil-lopreiato:
Delete it all!

nwalters512:
All that red makes me so happy :)

Issue #295, opened 2014-12-01
ListView->RecyclerView


phil-lopreiato:
All instances of ListView should be moved to [RecyclerView](https://developer.android.com/reference/android/support/v7/widget/RecyclerView.html), and the [ListViewAdapter](https://github.com/the-blue-alliance/the-blue-alliance-android/blob/master/android/src/main/java/com/thebluealliance/androidclient/adapters/ListViewAdapter.java) should move to extend [RecyclerView.Adapter](https://developer.android.com/reference/android/support/v7/widget/RecyclerView.Adapter.html)

PR #747, opened 2016-06-10
Moves team stats list to RecyclerView

nwalters512:
**Summary**: Team@Event stats list now uses RecyclerView. Woo!

**Issues Reference**: #295, for lack of anything better

**Test Plan**: Existing tests have been updated so that they still function

---
This change is [Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/747)


android/src/main/java/com/thebluealliance/androidclient/helpers/StatsHelper.java
```diff
@@ -0,0 +1,17 @@
+package com.thebluealliance.androidclient.helpers;
+
+import java.text.DecimalFormat;
+
+public final class StatsHelper {
```
phil-lopreiato: Do this in [`ThreadSafeFormatters`](https://github.com/the-blue-alliance/the-blue-alliance-android/blob/master/android/src/main/java/com/thebluealliance/androidclient/helpers/ThreadSafeFormatters.java) instead, so we can avoid race conditions

nwalters512: Good call, I forgot that was a thing

phil-lopreiato:
Reviewed 9 of 9 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/the-blue-alliance/the-blue-alliance-android/747)*

android/src/main/java/com/thebluealliance/androidclient/helpers/ThreadSafeFormatters.java
```diff
@@ -55,4 +55,8 @@ public static synchronized String formatDoubleOnePlace(double input) {
     public static synchronized String formatDoubleTwoPlaces(double input) {
         return DOUBLE_TWO_PLACES_FORMAT.format(input);
     }
+
+    public static synchronized String formatStat(double stat) {
```
phil-lopreiato: Nit: this is exactly the same as `formatDoubleTwoPlaces`, why not just use that method?

nwalters512: In case we want to change the formatting of just stats sometime in the future. I know that's unlikely, but I'd rather design for that now :stuck_out_tongue:

phil-lopreiato: It's an easy enough change to make in the future should we need it, I'd rather just keep simplicity in the code now. Plus, at the call side, `formatDoubleTwoPlaces` is more descriptive about what's happening than `formatStat` is

nwalters512: Fair enough. I guess my thinking is that we don't have to care what `formatStat` does as long as it's consistent and we use it for all stats. I'll make the change.

phil-lopreiato:
Reviewed 5 of 5 files at r2, 4 of 4 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/the-blue-alliance/the-blue-alliance-android/747)*

PR #752, opened 2016-06-16
Use actual padding for FAB padding instead of empty list item

nwalters512:
**Summary:** Instead of using an empty `Preference` to pad the bottom of the myTBA settings list, this now applies actual padding to the bottom of the list. This avoids the case of a random divider appearing in the list where it shouldn't (see screenshots).

**Issues Reference:** N/A

**Test Plan:** N/A

**Screenshots:**

Old (note the lone divider):

Imgur

New:

Imgur

---
This change is [Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/752)


phil-lopreiato:
:+1:

PR #754, opened 2016-06-17
[SQLite] Use quotes when matching team keys in JSON strings

nwalters512:
**Summary:** The previous LIKE statement never actually worked. Because the JSON strings we were searching looked like `...["frc111","frc254","frc1114"]...`, matching against something like `frc111,` didn't actually work because of the quotes. This PR matches against `"frc111"` instead. This also avoids `frc111` inadvertently matching `frc1114`.

**Issues Reference:** N/A

**Test Plan:** We really should find a way to test queries, because this has been broken for a long time and I only caught it because I happened to take a close look at this. That's beyond the scope of this PR, though.

---
This change is [Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/754)

### DIFF
--- a/android/src/main/java/com/thebluealliance/androidclient/database/tables/AwardsTable.java
+++ b/android/src/main/java/com/thebluealliance/androidclient/database/tables/AwardsTable.java
@@ -46,7 +46,7 @@ public class AwardsTable extends ModelTable<Award> {
 
     public List<Award> getTeamAtEventAwards(String teamKey, String eventKey) {
         Cursor cursor = mDb.rawQuery("SELECT * FROM `" + Database.TABLE_AWARDS + "` WHERE `" + EVENTKEY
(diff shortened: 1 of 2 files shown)


phil-lopreiato:
Nice find, this is why we can't have nice things.... >_<

I think we should be able to run `Database.getInstance()` with the Robolectric application context (I think it might not be super straightforward, but probably possible) to test some database interactions

phil-lopreiato:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/the-blue-alliance/the-blue-alliance-android/754)*

nwalters512:
We'd have to find a way to populate the database to test if queries work properly. We could just insert and then immediately query, but then query testing relies upon insertion code working properly. Which it should, but still... :/

Issue #706, opened 2016-04-25
[finally] Dynamic Tab Strips


phil-lopreiato:
Something we've wanted to do for a long time...

This way, we can only show tabs we have data for (so for events in the future, only show info & team list, for example).

If we [base the tab strip on RecyclerView](https://github.com/nshmura/RecyclerTabLayout), then we should be able to add/remove things from the adapter on the fly and just call `notifyDatasetChanged`. We'll have to make sure we don't cause weird `ViewPager` memory leaks though

nwalters512:
IIRC, the issue was dynamic tabs was with the pager adapter, not the tab strip. Adding/removing from a fragmentpageradapter is messy.

phil-lopreiato:
Ah, that does sound familiar, because `ViewPager` holds some references. I still can't believe this hasn't been made easier yet...

phil-lopreiato:
Continued from #741... (cc @1fish2)

> Workaround idea 1: Recreate the pager when it needs to add tabs. (Issue: Losing the current tab's contents.)

This will also recreate all the fragments involved (unless we wrote some way to save them). I don't think this is what we want, because it'll be a pretty big performance hit

> Workaround idea 2: Create the pager with spare tabs, hide them by limiting scrolling, and use them as needed.

You can't hide tabs either :(

The thing is, we'll have to find a way to update a `ViewPager` and its adapter dynamically. It's harder than you thing - google around for it and see if you can find any new solutions

nwalters512:
I have an inclination to just write a custom component that does what we want - although the fact that no one's already done that means it might not be too easy :/

Our next great library: Tabby McTabface?

phil-lopreiato:
If we do that project (you in the mood for a big summer refactor?) - I say we work towards ditching fragments entirely. See the "Deep in Fragment Spaghetti" section of [this article](https://corner.squareup.com/2014/10/advocating-against-android-fragments.html) and let's take a look at integrating [mortar](https://github.com/square/mortar) with our Subscriber/Binder framework

1fish2:
> This [recreating the pager when it needs to add tabs] will also recreate all the fragments involved (unless we wrote some way to save them). I don't think this is what we want, because it'll be a pretty big performance hit

[...]

PR #756, opened 2016-06-30
[WIP][Do Not Merge] SQLBrite integration

nwalters512:
**Summary:** Mostly making the PR to get feedback for what I've done so far. This all still needs _a lot_ of polishing.

The basic idea behind SQLBrite is that instead of manually requerying a database when data has changed, you simply subscribe to an observable that will automatically update whenever the query's associated tabled are changed. This allows us to decouple the process of loading data from the network and displaying it. A list of teams, for instance, just subscribes to an observable that will requery the database when the teams table is updated; we can then simply push new data into the database as the result of a network operation and the teams list will automatically update!

This is also useful for views that require data from multiple "sources" (e.g., EventInfoFragment relies on event details, event matches, team stats, and team rankings. Since all those things can be represented as reactive streams from the database, we can just merge those into a single stream of data that gets passed to the binder - no more shuttling data between components with the event bus!

So far I just converted two components: `EventInfoFragment` as a proof-of-concept for components that rely on data from multiple sources as mentioned above, and `EventTeamFragment` as a representative example of this applied to lists, which comprises most of our app.

I copied a bunch of components and prefixed them with `Brite` to differentiate them from the originals. This allows both the old and new data feeds to exist concurrently until we migrate everything.

**Issues Reference:** This was briefly discussed in #706

**Test Plan:** Largely utilizes existing codebase, so additionaly tests may or not be necessary. This should be determined later once things are closer to their final state.

---

This change is [Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/756)

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -289,6 +289,10 @@ dependencies {
     testCompile "org.robolectric:shadows-support-v4:3.0"
     testCompile "org.robolectric:shadows-multidex:3.0"
     testCompile 'junit:junit:4.12'
-    testCompile 'org.mockito:mockito-core:2.0.31-beta'
+    //testCompile 'org.mockito:mockito-core:2.0.31-beta'
+    testCompile "org.powermock:powermock-module-junit4:1.6.4"
+    testCompile "org.powermock:powermock-module-junit4-rule:1.6.4"
+    testCompile "org.powermock:powermock-api-mockito:1.6.4"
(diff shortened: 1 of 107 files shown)


phil-lopreiato:
After a quick skim, I'm a fan! I really like having the datafeed have a "one way" flow, hopefully it makes things much simpler and easier to understand.

PR #761, opened 2016-07-15
Logging Consolidation

phil-lopreiato:
**Summary:** Integrate Firebase Crash Reporting and refactors logging classes. This diff creates a custom logging class that wraps `android.util.Log`. It automatically sets the tag (so we can kill `Constants.LOG_TAG`!) and reports the items to Firebase. This has to manually stop `DEBUG` and `VERBOSE` items from being printed in prod builds.

**Test Plan:** **DO NOT MERGE YET** This won't work for CI builds and for people who haven't configured Firebase (specifically, who don't have a proper `google-services.json`). If only Firebase made actually using crash reporting optional >_<

---
This change is [Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/761)


phil-lopreiato:
I've updated this to just refactor our log statements into one TBA class that we can extend in the future

Issue #780, opened 2016-10-05
Tap a summary notification -> "The Blue Alliance Debug has stopped"


1fish2:
Tapping a summary notification sometimes makes the app exit summarily. The OS displays a dialog "The Blue Alliance Debug has stopped" over the home screen.

There's no acknowledgement of the crash in logcat (!) even with filters turned off but Android Studio's Run view displays a stack traceback. The stack has no TBA code but it does have third party library code `io.nlopez.smartadapters`.

This occurs in the current master branch and also in the branch [issue_774](https://github.com/1fish2/the-blue-alliance-android/tree/issue_774) where I'm testing a fix for #774.

```
W/GAv4: Discarding hit. Missing tracking id parameter: a=1869011076,an=The Blue Alliance Debug,exd=NullPointerException (@DefaultBindableLayoutBuilder:viewType:38) {main},t=exception,sf=100.0,tid=,exf=1,aid=3.4.8-master.2+gd52ffe5,av=3.4.8-master.2+gd52ffe5
W/GAv4: syncDispatchLocalHits timed out: java.util.concurrent.TimeoutException
E/AndroidRuntime: FATAL EXCEPTION: main
                  Process: com.thebluealliance.androidclient.development, PID: 3720
                  java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.Class java.lang.Object.getClass()' on a null object reference
                      at io.nlopez.smartadapters.builders.DefaultBindableLayoutBuilder.viewType(DefaultBindableLayoutBuilder.java:38)
                      at io.nlopez.smartadapters.adapters.RecyclerMultiAdapter.getItemViewType(RecyclerMultiAdapter.java:134)
                      at android.support.v7.widget.RecyclerView$Recycler.getViewForPosition(RecyclerView.java:4664)
                      at android.support.v7.widget.RecyclerView$Recycler.getViewForPosition(RecyclerView.java:4617)
                      at android.support.v7.widget.LinearLayoutManager$LayoutState.next(LinearLayoutManager.java:1994)
                      at android.support.v7.widget.LinearLayoutManager.layoutChunk(LinearLayoutManager.java:1390)
                      at android.support.v7.widget.LinearLayoutManager.fill(LinearLayoutManager.java:1353)
                      at android.support.v7.widget.LinearLayoutManager.onLayoutChildren(LinearLayoutManager.java:574)
[...]

Issue #774, opened 2016-09-25
[Nougat] Notification Summary Broken


1fish2:
At Chezy Champs 2016 today, match scores finally started flowing through to TBA at 19:45 PT, after the event ended for the day. Then match score Firebase Cloud Messages (FCMs) started arriving every few seconds. I watched them arrive on a Nexus 7 Marshmallow tablet and a Nexus 5X Nougat phone.

When I pulled down an FCM's notification view and let the view stay open, when following FCMs arrived, one or both devices played notification tones, Marshmallow accumulated additional lines in the app's notification view, but Nougat did not.

When I swiped away the notification view on Nougat, those additional match-score notification lines never did appear although they are present in the app's Recent Notifications list, and the next incoming FCM would appear as a new notification view.

Apparently, additional info arriving while the app's pull-down notification view is open doesn't successfully add to that view.

1fish2:
Update: This Nougat bug happens even when the pull-down notification is closed. With an un-dismissed notification, more arriving FCMs don't add to the notification view.

phil-lopreiato:
Yay compatibility bugs (this one falls under #663)... Thanks for reporting!

1fish2:
Using Nougat's bundled notifications would be nice, but `InboxStyle` notifications should still work.

Go ahead and assign this one to me and I'll look into it this week.

phil-lopreiato:
Awesome, thanks!

I'd guess the bugs starts with how we auto-dismiss the first notification before replacing it?

I believe the group summary is backwards compatible, so it may be worth it to just migrate the whole thing to N notifications and let the fallback do its thing.

1fish2:
It should arrange for the new notification to replace any previous one, although offhand I forget how that's done. Maybe it shouldn't call `setGroupSummary()` since the app does its own summarization. Moving to N notifications is a fine idea.

phil-lopreiato:
> Maybe it shouldn't call setGroupSummary() since the app does its own summarization. Moving to N notifications is a fine idea.

I think we should let the system handle all the summaries instead of doing it ourselves (assuming it's backwards compatible).

1fish2:
> I think we should let the system handle all the summaries instead of doing it ourselves (assuming it's backwards compatible).

Yes, I like that a lot but I'd be more I'm comfortable first checking a couple things so we're less likely to leave a bug in the code (e.g. calling `setGroupSummary()`?) that could come back to haunt us.

[...]

PR #781, opened 2016-10-06
Fix Issue 774 and implement stacked notifications

1fish2:
**Summary:** 
* Fix broken notification groups on Android N.
* Implement stacked (grouped, bundled) notifications for Android N and Watch. The technique is to post the component notifications then a group summary notification. See [Stacking Notifications](https://developer.android.com/training/wearables/notifications/stacks.html).
* Use Notification[Manager]Compat from the latest support lib version to get their compatibility workarounds. That requires compileSdkVersion 24.
* Workaround notification-grouping bugs on API 20 (K) and 21 (L pre-MR1).
* Rephrase some notification content text messages to put the useful info
up front so it’s readable when collapsed to one line and even in the
summary view that shows each entry’s title + content in a single line.

Note that on N, tapping or swiping one notification in a stack of them dismisses just that one. That's useful and the design intention, but to dismiss many accumulated notifications at once, you'll have to collapse them or Clear All notifications.

**Issues Reference:** 

* #774: [Nougat] Notification Summary Broken 
* #780: tapping the content of a summary notification sometimes crashes, but that's independent of this CL.
* Tapping a match-score notification doesn't always open a relevant TBA screen. How to reproduce that bug? 
* The app's DB can have leftover FCM notifications but no notification shade entries, e.g. after installing an upgrade. (Also after rebooting?) The next received FCM notification will display a Summary notification that includes the leftover notifications + the new one. If it only happens after upgrading, the app could watch for a MY_PACKAGE_REPLACED intent.

Q. Should we remove the large icons from these notifications on N to make more room for the content text? See the screenshots.

**Test Plan:** 
- [ ] Android versions [16 .. 19] -- on Android ≤ 21, the app just shows one notification at a time, either a component notification or a summary notification
- [x] Android versions [20 .. 24]
- [x] without Wear
- [ ] with Wear
- [x] with notification shade {empty, some closed ones, a collapsed one, an expanded one, a collapsed summary, an expanded summary}
- [x] on M and N with `targetSdkVersion` temporarily set to 23 and 24
- [x] on N with heads-up notifications enabled
- [ ] on N with heads-up notifications disabled

**Screenshots:**
Stacked notifications on Android N [Here, the large icons take space away from the content text.]:
n stacked notifications medium

the user can further expand any of the component notifications:
n stacked notifications large

collapsed:
n stacked notifications small

Android M just shows the same-old summary notification:
m summary notification

---
This change is [Reviewable](https://reviewable.io/reviews/the-blue-alliance/the-blue-alliance-android/781)


android/build.gradle
```diff
@@ -37,8 +37,8 @@ project.afterEvaluate {
 }
 
 android {
-    compileSdkVersion 23
-    buildToolsVersion '23.0.2'
+    compileSdkVersion 24
```
phil-lopreiato: Can you update the versions in `.travis.yaml` also?

1fish2: I updated `.travis.yml`. Let me know if there's any problem with that. Hopefully Travis will built it properly now.

(I left `com.android.tools.build:gradle:2.1.3` since the command-line gradlew couldn't load build tools 2.2.0 in my BBQ Timer app. Maven central doesn't have it yet?)

I'll remove the large icons from the notifications on N.

1fish2: Done.

nwalters512:
This looks awesome! I second removing the large icons, probably on all platforms. The notification titles are much more informative of the content of the notification and at worst only slightly more difficult to parse than the icons.

1fish2:
Removing the large icons from all platforms is simpler. I'll do that.

I was about to make a subclass `NotificationCompat.Builder` to centralize the OS-version-specific handling that the Compat library doesn't finish. `setLargeIcon()` and `setGroup()` would no-op on N+, and `setGroupSummary(true)` would limit the priority to `PRIORITY_DEFAULT`. I'll skip that for now.

fangeugene:
While you're getting rid of the large icons, can you make sure the lamp is blue instead of rust gold?

1fish2:
Yes! It looks like primary is the right blue color resource to use. LMK if there's a better one -- this one doesn't match the colors in `ic_launcher`.

1fish2:
I set it to use the `primary` color, which is red in debug builds. Definitely more fitting than gold.

That makes the group look like this on N without the large icons. Strangely, `SummaryNotification` didn't set a color, defaulting to gray. Presumably because that's a different color than the component notifications, all the small icons appear:

n all notifications with small icons 1

By setting the `SummaryNotification` small icon color consistent with the component notifications, the appearance got simpler:

n all notifications with consistent small icons 1

On M, a component notification looks like this without the large icon:
m one notification 1

and the group looks like this:
m all notifications 1

1fish2:
These screenshots show:

* Notifications on N automatically show the app name, in red here. Notifications on M- do not.
* `SummaryNotification` calls `setSummaryText(app_name)`. On N that shows a redundant "The Blue Alliance Debug" to the right of the red app name. On M- that shows the app name at the bottom, below the InboxStyle entries and the horizontal rule. An email app might name the email account there.

Consider:
* On M-, we could put "and 2 more..." or "9 new notifications" in the summary line instead of "The Blue Alliance Debug", leaving room for one more entry in the list.
* On N+, we could make the summary show empty or "9 new notifications".

phil-lopreiato:
Awesome sauce!

> On M-, we could put "and 2 more..." or "9 new notifications" in the summary line instead of "The Blue Alliance Debug", leaving room for one more entry in the list.

This is a good idea

> On N+, we could make the summary show empty or "9 new notifications".

I think "9 new notifications" would be best, but that might be tricky to keep in sync if a single notification is dismissed. I'm fine with just setting it to be blank.

1fish2:
I made those changes and I have a hypothesis on #774: [Nougat] Notification Summary Broken. If that pans out, I'll put the fix in a separate branch.

1fish2:
I meant the #780 crash. Edited the previous comment.

1fish2:
Travis CI: "Gradle build daemon disappeared unexpectedly"?

phil-lopreiato:
I reran the build and everything is okay. Is this good to ship?

1fish2:
This is good to ship although I forgot to post yesterday's screenshots.

Here's the summary on M up to 7 component notifications now:
m up to 7 1

and the overflow:
m 7 1 1

and the overflow case on N, where Android handles the "+1" part, even changing it dynamically when a new heads-up notification times out and merges into the summary:
n 6 notifications 1

phil-lopreiato:
Perfect, thanks a ton!

PR #788, opened 2016-10-28
Track Last-Modified Times in DB

### LINKED ISSUE #638: “No alliance data found” after receiving “Alliance selections updated” notification

phil-lopreiato:
**Summary:** We can now track when a model is last modified (from the `Last-Modified` header) and only do writes when we have newer data. However, the way last modified added is hacky, due to the way is has to be added (since it has to come from the `Response` object), so it shouldn't be used other than for db writes.

**Issues Reference:** Should fix #638 

**Test Plan:** Didn't break unit tests, need to add more and do lots of manual testing


1fish2:
Yay!!! I didn't read the code yet but this is top item on my wish list.

Q. Is a network request's Last-Modified timestamp guaranteed to be right for all DB rows that come from it? I.e., are they safe from version skew in all cache layers including receipt from FMS?

Q. Can we get Last-Modified timestamps in push notifications yet?

phil-lopreiato:
> Q. Is a network request's Last-Modified timestamp guaranteed to be right for all DB rows that come from it?

Yeah, the timestamps are generated in App Engine, so there shouldn't be any skew there (all comparisons are done with those timestamps)

> Q. Can we get Last-Modified timestamps in push notifications yet?

Yeah, that's not hard, they just have to be added

1fish2:
Ah. Although the data in one API reply must've been _"last modified by"_ FMS at multiple times, if the client uses the `Last-Modified` timestamp for a _"current-as-of"_ timestamp, it'll prevent retrograde motion even though it won't prevent redundant DB writes. That's fine.

Maybe name that column something like `timestamp` rather than `last_modified`.

PR #789, opened 2016-10-29
[DO NOT MERGE] Use FCM for Push Notifications

phil-lopreiato:
**Summary:** This is a rough skeleton of migrating our old GCM code to use the new FCM libraries. 

**DO NOT MERGE**: This currently fails with `FirebaseMessaging: Received message with unknown type: X` when an incoming message is received.


bherbst:
This was done in https://github.com/the-blue-alliance/the-blue-alliance-android/pull/955

PR #838, opened 2017-08-12
Fix incorrect LeakCanary initialization

### LINKED ISSUE #829: fatal error in leakcanary in debug build because FirebaseApp is not initialized
This crash only happens in debug builds, right?

```
03-15 00:37:46.608 15802-15802/? E/AndroidRuntime: FATAL EXCEPTION: main
                                                   Process: com.thebluealliance.androidclient.development:leakcanary, PID: 15802
                                                   java.lang.RuntimeException: Unable to create application com.thebluealliance.androidclient.TBAAndroid: java.lang.IllegalStateException: Default FirebaseApp is not initialized in this process com.thebluealliance.androidclient.development:leakcanary. Make sure to call FirebaseApp.initializeApp(Context) first.
                                                       at android.app.ActivityThread.handleBindApplication(ActivityThread.java:5406)
                                                       at android.app.ActivityThread.-wrap2(ActivityThread.java)
                                                       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1545)
                                                       at android.os.Handler.dispatchMessage(Handler.java:102)
                                                       at android.os.Looper.loop(Looper.java:154)
                                                       at android.app.ActivityThread.main(ActivityThread.java:6119)
                                                       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886)
                                                       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776)
                                                    Caused by: java.lang.IllegalStateException: Default FirebaseApp is not initialized in this process com.thebluealliance.androidclient.development:leakcanary. Make sure to call FirebaseApp.initializeApp(Context) first.
                                                       at com.google.firebase.FirebaseApp.getInstance(Unknown Source)
                                                       at com.google.firebase.crash.FirebaseCrash.zzcqj(Unknown Source)
                                                       at com.google.firebase.crash.FirebaseCrash.log(Unknown Source)
                                                       at com.thebluealliance.androidclient.TbaLogger.i(TbaLogger.java:46)
                                                       at com.thebluealliance.androidclient.TBAAndroid.onCreate(TBAAndroid.java:62)
                                                       at android.app.Instrumentation.callApplicationOnCreate(Instrumentation.java:1024)
                                                       at android.app.ActivityThread.handleBindApplication(ActivityThread.java:5403)
                                                       at android.app.ActivityThread.-wrap2(ActivityThread.java) 
                                                       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1545) 
                                                       at android.os.Handler.dispatchMessage(Handler.java:102) 
                                                       at android.os.Looper.loop(Looper.java:154) 
                                                       at android.app.ActivityThread.main(ActivityThread.java:6119) 
                                                       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886) 
                                                       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776) 
```

SUPERCILEX:
Fixes #829

### DIFF
--- a/android/src/main/java/com/thebluealliance/androidclient/TBAAndroid.java
+++ b/android/src/main/java/com/thebluealliance/androidclient/TBAAndroid.java
@@ -59,6 +59,9 @@ public class TBAAndroid extends MultiDexApplication {
     @Override
     public void onCreate() {
         super.onCreate();
+        if (LeakCanary.isInAnalyzerProcess(this)) return;
+        else LeakCanary.install(this);
+
         TbaLogger.i("Welcome to The Blue Alliance for Android, v" + BuildConfig.VERSION_NAME);
         getDatafeedComponenet().inject(this);
         mAppConfig.updateRemoteData();
@@ -71,10 +74,6 @@ public class TBAAndroid extends MultiDexApplication {
                             .enableWebKitInspector(Stetho.defaultInspectorModulesProvider(this))
 
                             .build());
-        }
-
-        if (Utilities.isDebuggable()) {
-            LeakCanary.install(this);
         }
     }
 
--- a/android/src/main/java/com/thebluealliance/androidclient/TBAAndroid.java
+++ b/android/src/main/java/com/thebluealliance/androidclient/TBAAndroid.java
@@ -59,8 +59,12 @@ public class TBAAndroid extends MultiDexApplication {
     @Override
     public void onCreate() {
         super.onCreate();
-        if (LeakCanary.isInAnalyzerProcess(this)) return;
-        else LeakCanary.install(this);
+        if (LeakCanary.isInAnalyzerProcess(this)) {
+            // This process is dedicated to LeakCanary for heap analysis.
+            // You should not init your app in this process.
+            return;
+        }
+        LeakCanary.install(this);
 
         TbaLogger.i("Welcome to The Blue Alliance for Android, v" + BuildConfig.VERSION_NAME);
         getDatafeedComponenet().inject(this);


android/src/main/java/com/thebluealliance/androidclient/TBAAndroid.java
```diff
@@ -59,6 +59,9 @@ public TBAAndroid() {
     @Override
     public void onCreate() {
         super.onCreate();
+        if (LeakCanary.isInAnalyzerProcess(this)) return;
```
1fish2: Thanks for the fix!

I recommend using braces around each "then" and "else" branch except maybe some one-liners like this line. Even when formatting on one line, I usually write explicit braces. Otherwise it's easy to accidentally break code e.g. by adding a Log.d() call.

Shall we copy in the explanatory comment from [LeakCanary README](https://github.com/square/leakcanary)?
```
      // This process is dedicated to LeakCanary for heap analysis.
      // You should not init your app in this process.
```

After then-return, `else` is confusing; the `LeakCanary.install(this)` call is effectively in the same block as the following code.

phil-lopreiato: Agree on both of @1fish2's comments. The formatting should look like:

```java
public void onCreate() {
    super.onCreate();
    if (LeakCanary.isInAnalyzerProcess(this)) {
        // Some descriptive comment about why this is necessary
        return;
    }
    LeakCanary.install(this);
    ...
}
```

SUPERCILEX: Yeah, it was late and I was being lazy 😄

SUPERCILEX:
@phil-lopreiato Any updates on this? `master` seems to be broken without my help so this should be good to go. 😄

phil-lopreiato:
Yeah, this is fine. I need to exclude some files from the checkstyle runs (which is why master is broken)

Issue #842, opened 2017-11-11
NPE when a team has no name


1fish2:
The 9000 series teams at MadTown Throwdown today lack team names. This causes at least one crash:

```
11-10 23:17:24.816  6133  6133 E AndroidRuntime: Process: com.thebluealliance.androidclient, PID: 6133
11-10 23:17:24.816  6133  6133 E AndroidRuntime: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.isEmpty()' on a null object reference
11-10 23:17:24.816  6133  6133 E AndroidRuntime:    at com.thebluealliance.androidclient.listitems.StatsListElement.getView(StatsListElement.java:45)
11-10 23:17:24.816  6133  6133 E AndroidRuntime:    at com.thebluealliance.androidclient.adapters.ListViewAdapter.getView(ListViewAdapter.java:29)
11-10 23:17:24.816  6133  6133 E AndroidRuntime:    at android.widget.AbsListView.obtainView(AbsListView.java:2372)
...
```

I'll send a PR.

phil-lopreiato:
Nice find! I also updated APIv3 to provide a default value (since I'm pretty sure we treat team names as required), so it'll be best to have this fixed in both places

PR #843, opened 2017-11-11
Fix bug #842

1fish2:
**Summary:** 

At MadTown 2017, the 2nd and 3rd robots have 8xxx and 9xxx numbers but
no names.

```
java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.isEmpty()' on a null object reference
  at com.thebluealliance.androidclient.listitems.StatsListElement.getView(StatsListElement.java:45)
```

**Issues Reference:** 

#842

**Test Plan:** 

Manually tested.

**Screenshots:**

### DIFF
--- a/android/src/androidTest/java/com/thebluealliance/androidclient/listitems/StatsListElementTest.java
+++ b/android/src/androidTest/java/com/thebluealliance/androidclient/listitems/StatsListElementTest.java
@@ -1,16 +1,18 @@
 package com.thebluealliance.androidclient.listitems;
 
+import android.content.Context;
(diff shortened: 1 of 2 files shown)


1fish2:
The build passed before I added a test case.

The added test case works fine in AS as long as I run it on Android L. Apparently Screenshot tests don't work on M+.

How does one view the Travis test report?

phil-lopreiato:
You can click on the "X" to see the job details. I re-ran it and it passed, so I guess that's good.

Anyway, thanks for the fix!

Issue #872, opened 2018-09-16
Crash due to Invalid district key "2018tx"


1fish2:
On Android 8.1 emulator with Google Play (also reported on 8.0), do this:

1. Open the "Districts" screen via the hamburger menu.
2. Scroll down.
3. Observe crash due to invalid district key `"2018tx"`.

The problem is the string `"tx"` is not in the list of strings in `DistrictType.fromAbbreviation()`.

```
    --------- beginning of crash
09-15 21:19:22.197 4387-4387/com.thebluealliance.androidclient.development E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.thebluealliance.androidclient.development, PID: 4387
    java.lang.IllegalArgumentException: Invalid district key
        at com.thebluealliance.androidclient.listeners.DistrictClickListener.<init>(DistrictClickListener.java:21)
        at com.thebluealliance.androidclient.listitems.DistrictListElement.getView(DistrictListElement.java:61)
        at com.thebluealliance.androidclient.adapters.ListViewAdapter.getView(ListViewAdapter.java:29)
        at android.widget.AbsListView.obtainView(AbsListView.java:2365)
        at android.widget.ListView.makeAndAddView(ListView.java:2052)
        at android.widget.ListView.fillDown(ListView.java:786)
        at android.widget.ListView.fillGap(ListView.java:750)
        at android.widget.AbsListView.trackMotionScroll(AbsListView.java:5225)
        at android.widget.ListView.trackMotionScroll(ListView.java:1971)
        at android.widget.AbsListView$FlingRunnable.run(AbsListView.java:4769)
        at android.view.Choreographer$CallbackRecord.run(Choreographer.java:911)
        at android.view.Choreographer.doCallbacks(Choreographer.java:723)
        at android.view.Choreographer.doFrame(Choreographer.java:655)
        at android.view.Choreographer$FrameDisplayEventReceiver.run(Choreographer.java:897)
        at android.os.Handler.handleCallback(Handler.java:790)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:164)
        at android.app.ActivityThread.main(ActivityThread.java:6494)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:438)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:807)
[...]

PR #861, opened 2018-03-30
Prereqs for further development; a more thematic notification sound

1fish2:
**Summary:** 

* Update to current tools (without changing targetApi yet): Android Studio 3.1, build tools 27.0.3, gradle 4.4, android gradle 3.1.0, current repositories, compileSDK 26, JDK 1.8, remove Retrolambda, Android gradle DSL updates, newer support libs.
* Update to a more thematic notification sound It’s also higher pitch so it should be easier to hear from phone speakers.
* Don't show OPRs before there are any match stats.

**Issues Reference:** 

**Test Plan:** 
It works on my phone.

**Screenshots:**
device-2018-03-29-222458

PR #873, opened 2018-09-16
Minimal fix for crash #872

1fish2:
**Summary:** 

Scrolling down the 2018 and 2019 Districts screen to where "FIRST In Texas District" should appear will crash.

**Issues Reference:** 
#872 

**Test Plan:** 
Manual test.

**Screenshots:**
screenshot_1537073111

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
-#Fri Sep 16 18:38:07 EDT 2016
+#Sat Sep 15 19:45:22 PDT 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-3.3-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-all.zip
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
-#Sat Sep 15 19:45:22 PDT 2018
+#Sat Sep 22 20:44:25 PDT 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.10-all.zip
--- a/libImgur/build.gradle
+++ b/libImgur/build.gradle
@@ -4,6 +4,6 @@ sourceCompatibility = 1.6
 targetCompatibility = 1.6
 
 dependencies {
-    compile "com.squareup.retrofit2:retrofit:${retrofitVersion}"
-    compile "io.reactivex:rxjava:${rxJavaVersion}"
+    implementation "com.squareup.retrofit2:retrofit:${retrofitVersion}"
+    implementation "io.reactivex:rxjava:${rxJavaVersion}"
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,12 +9,9 @@ android:
   components:
     - platform-tools
     - tools
-    - build-tools-27.0.3
+    - build-tools-28.0.2
     # The SDK version used to compile the project
     - android-26
-    - android-25
-    - android-24
-    - android-23
     - extra-android-m2repository
     - extra-google-m2repository
     - extra-google-google_play_services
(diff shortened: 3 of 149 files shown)


1fish2:
This PR does not try to update travis-ci nor fully fix the build files. See PR #861 for a broader attempt, if not 100% successful.

1fish2:
Oops, I thought I pushed the whole fix the other day.

That's the minimal fix. One could go a little further and log an unrecognized district code. Also of course update the travis CI build. I could take a guess at that but it's a mystery to me.

sorianog:
@1fish2 You still planning on going forward with this PR? Looks like Travis ain't happy 😢

sorianog:
@1fish2 Hmm, I can take a look at the tests after you're done.

1fish2:
@sorianog I pushed a cut at making travis-ci work. Please do look at the tests. Many thanks!

We must push a release before Nov. 1. Starting Nov. 1, releases must support API 26. That's where Robolectric, library versions, and Android changes hit the fan.

sorianog:
@1fish2 Well, I'm glad I'm here to help since what you explained is definitely something I helped work on ("app currency" - keep app updated with latest libs/tools/frameworks) at my full-time gig. Esp, Roboelectric stuff. Want to be careful with the new setup for injections and what not. I'll keep that deadline in mind!

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Fri Sep 16 18:38:07 EDT 2016
+#Sat Sep 15 19:45:22 PDT 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-3.3-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-all.zip
```
sorianog: Ah, good this is a def a req for increasing your API level support. Gradle needs to be updated as you did here!

libImgur/build.gradle
```diff
@@ -4,6 +4,6 @@ sourceCompatibility = 1.6
 targetCompatibility = 1.6
 
 dependencies {
-    compile "com.squareup.retrofit2:retrofit:${retrofitVersion}"
-    compile "io.reactivex:rxjava:${rxJavaVersion}"
-}
\ No newline at end of file
+    implementation "com.squareup.retrofit2:retrofit:${retrofitVersion}"
+    implementation "io.reactivex:rxjava:${rxJavaVersion}"
```
sorianog: I like how vars are used here to dynamically change the version of chosen libs/frameworks. Is this something that can be done with more of the libs in this project?

gradle/wrapper/gradle-wrapper.properties
1fish2: Terrific! Welcome!

I have another cut at updating for Robolectric compatibility. I'll push that shortly after trying further to work through the errors that arise at each attempt.

Their [migration guide](http://robolectric.org/migrating/) is somewhat helpful. The good news is they've updated Robolectric for current or almost-current Android APIs. The bad news is various parts of their APIs changed and various other tools and libs have to upgrade to get there.

In general, the Android gradle rules (esp. for this project) and library dependencies are a mystery to me. Docs say we need to turn on `includeAndroidResources = true` but does that mean we should turn off the `build.workaround-missing-resource.gradle`?

libImgur/build.gradle
1fish2: Phil put this in. I surmise that it's particularly useful for the nested projects libImgur and libTba. I don't know why those are nested projects.

The build still gets into trouble when dependent libraries use different versions of these dependencies, not to mention outdated gradle commands like `compile` and task `<<`.

1fish2: It's particularly useful with `${supportLibVersion}`. It does add complexity and surprises, like when gradle complains that we're not using the same support library version everywhere (which must be due to an indirect dependency), so I'd want to extend it only to places where the same version number must appear in multiple places.

1fish2: Do the current build tools make retrofit obsolete?

1fish2:
@sorianog I'm out of ideas to try other than updating library versions, which might have n^2 compatibility problems but it might fix some of the build problems like `Configuration 'compile' is obsolete`. So if you can make progress, please do.

sorianog:
@1fish2 Thanks for the work thus far! I'll see what I can do.

libImgur/build.gradle
sorianog: @1fish2 Nah, I think retrofit is still okay to include. That's how it is at my work's codebase at least!

phil-lopreiato:
I removed the hard-coded districts entirely in https://github.com/the-blue-alliance/the-blue-alliance-android/pull/874 so I'm going to close this, but let's definitely figure out all the inane things we need to do to get the app to build with new tools

Issue #868, opened 2018-05-17
Scrolling to the bottom of 2018 districts causes a crash


randomtestfive:
It appears to be an issue with validating the FIRST in Texas District.
Looks like Texas is missing from the DistrictType enum.

miklast:
Still an issue.

1fish2:
Duplicated by #872 and a proposed fix in PR #873.

PR #874, opened 2018-10-02
Rip out hard coded districts

phil-lopreiato:
**Summary:** 
Keeping a hard-coded list is stupid and I should have ripped all this out years ago

**Issues Reference:** 
Fixes https://github.com/the-blue-alliance/the-blue-alliance-android/issues/872
Fixes https://github.com/the-blue-alliance/the-blue-alliance-android/issues/868

**Test Plan:** 
ran it

**Screenshots:**
42879973_175725303307884_5088016845244465152_n


android/src/main/java/com/thebluealliance/androidclient/database/Database.java
```diff
@@ -128,7 +128,6 @@
             + DistrictsTable.KEY + " TEXT PRIMARY KEY NOT NULL, "
             + DistrictsTable.ABBREV + " TEXT NOT NULL, "
             + DistrictsTable.YEAR + " INTEGER NOT NULL, "
-            + DistrictsTable.ENUM + " INTEGER NOT NULL,"
```
1fish2: Any compatibility problems with existing tables in already-installed apps?

android/src/main/java/com/thebluealliance/androidclient/helpers/DistrictHelper.java
```diff
@@ -1,25 +1,15 @@
...
-        }
-        try {
-            Integer.parseInt(key.substring(0, 4));
-            String districtAbbrev = key.substring(4);
-            return DistrictType.fromAbbreviation(districtAbbrev) != DistrictType.NO_DISTRICT;
-        } catch (NumberFormatException e) {
-            return false;
-        }
+        if (key == null || key.isEmpty()) return false;
+        return key.matches("^[1-9]\\d{3}[a-z,0-9]+$");
```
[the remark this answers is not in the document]
phil-lopreiato: Ugh, I just copied this from the event one. It probably is valid, even if it never happens, we don't actually enforce that many restrictions on them

android/src/main/java/com/thebluealliance/androidclient/database/Database.java
phil-lopreiato: Shouldn't be - this just leaves the data hanging (not a huge deal since it's pretty small)

PR #887, opened 2019-02-09
Fix some offseason events not showing up

### LINKED ISSUE #870: Some off-season events are missing from the list, maybe because they straddle a month boundary
Steps to reproduce:

1. Search the app for `texas robotics invitational`.
2. Result: 5 events found: Jun 21, 2014; Jun 20, 2015; Jun 24 to Jun 25, 2016; **Jul 1, 2017**; Jun 29 to Jun 30, 2018.
1. Search the app for `Qianjiang`.
2. Result: 1 event found: **Jul 29 to Aug 2, 2018**.
3. Look in the menus for 2017 - JUN OFFSEASON EVENTS, 2017 - JUL OFFSEASON EVENTS, 2018 - JUL OFFSEASON EVENTS, 2018 - AUG OFFSEASON EVENTS.
4. Result: There are events [assuming you pushed the app to refresh its stale fragment menus -- do we still have an open Issue for that bug?], but the two bold events, above aren't visible in those lists.

**Hypothesis:** These are off-season events that straddle month boundaries and get filtered out of all months. (Rohit Gondi points out that TRI straddled June - July, 2017.)

bherbst:
**Summary:** 
The logic for fetching events within a given month was funky in a few ways:
 * It only kept events that were completely within the month (`monthStart <= eventStart && eventEnd < eventEnd`)
 * This also actually would have excluded events ending on the last day of the month
 * Both the month start and the month end were calculated with the correct year, month, and day, but using the current device time. This effectively meant that unless you were getting the events at precisely midnight, you wouldn't see any events starting on the first of the month. 

This new logic uses the correct time, and only checks that the event's start date lies within the month. If an event starts in July and ends in August, it is bucketed as a July event.

**Issues Reference:** 
Fixes #870 

**Test Plan:** 
Unfortunately the APICache (particularly with the database) isn't particularly well set up for unit tests, but I'd like to get there eventually.

For manual testing, September 2018 contains events on September 1st (Battleship Blast Day 1) and September 30th (2018 CORI Invitational) that serve as good start/end logic checks within the month.
July 2018 contains an event (RCC aka Region China Champions) that straddles the July - August boundary.

Battleship Blast Day 1 and RCC previously did not show up, and now do.

### DIFF
--- a/android/src/main/java/com/thebluealliance/androidclient/datafeed/APICache.java
+++ b/android/src/main/java/com/thebluealliance/androidclient/datafeed/APICache.java
@@ -203,6 +203,7 @@ public class APICache {
         return Observable.create((observer) -> {
             try {
                 Calendar cal = Calendar.getInstance();
+                cal.clear();
                 cal.set(year, month, 1);
                 String start = Long.toString(cal.getTimeInMillis());
                 cal.add(Calendar.MONTH, 1);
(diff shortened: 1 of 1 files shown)


phil-lopreiato:
Nice catch! You're right - this was super clowny the way it was before, thanks for the fix!

PR #893, opened 2019-02-28
Add auth module to imgur suggestion service and GCM message handler

bherbst:
**Summary:** 
We weren't creating an `AuthModule` (see #884, also related #891) for the Imgur suggestion service and the GCM message handler components, so both were crashing.

This resolves both issues.

**Test Plan:** 
I'm not set up to test GCM notifications at the moment, but sending an FCM message from the Firebase console, or launching either of these services through other means all result in the same crash. After this fix, it does not crash.


phil-lopreiato:
@bherbst :fire:

PR #917, opened 2019-04-19
[notification] Don't use message_type key in payloads

phil-lopreiato:
Long story.

Mainly, this is a reserved key in FCM that wasn't in GCM, so we see random failures.


bherbst:
👍

Issue #911, opened 2019-03-21
Migrate to AndroidX


bherbst:
We should migrate to AndroidX, the future of the support libraries. The existing support libs will not be moving past 28.

Since we are already on v28 of the support lib, the migration shouldn't be too bad.

bherbst:
Completed in #919

PR #919, opened 2019-05-22
Migrate to AndroidX (#911)

bherbst:
**Summary:** 
The support libraries are now collectively part of AndroidX. This initial work uses the drop-in replacements for support library v28 (AndroidX 1.0), but there are newer versions of AndroidX packages that contain work (bugfixes, features, etc) not reflected in the support libraries.

By getting us on Android X, this PR enables us to start using the newer libraries in the future.

As much as possible, I tried to contain the changes to the minimum, but there were a few cascading issues that required me to touch more than I wanted.

 * I had to update us to a newer version of Gradle, which removed some deprecated APIs and consequently required updating a few plugins
 - * The "missing resources workaround" stuff is no longer needed, and wasn't compatible, so I removed
 - * The screenshot plugin needed to be updated 
 - * Robolectric needed to be updated
 - * Jacoco needed to be updated

The Robolectric update was interesting in that for some reason it started to throw `NoClassDefFound` errors on our `TestTbaAndroid` class. Renaming the `TBAAndroid` application class to match the casing of `TestTbaAndroid` fixed that, which caused a lot more files to change than I wanted.

The vast majority of these changes are simply replacing the old package names with the new ones, and the vast majority of those changes were done with the automatic refactoring built into Android Studio.

Highlights of meaningful changes:
 * gradle.properties
 * build.gradle
 * android/build.gradle
 * code_coverage.gradle
 * spectrum/build.gradle

**Issues Reference:** 
#911 

**Test Plan:** 
Yeesh. I ran the unit tests and poked around the app for a while, but the more people using it prior to release the better.


bherbst:
🤞 for Travis

bherbst:
At least one flaky test is making Travis sad, but I've seen each step succeed individually 🙃

phil-lopreiato:
:open_mouth: 

This is awesome!

PR #929, opened 2019-11-22
Migrate team list to ViewPager2

### LINKED ISSUE #706: [finally] Dynamic Tab Strips

bherbst:
**Summary:** 
ViewPager2 has been released! ViewPager2 is a re-write of ViewPager that now leverages `RecyclerView` to manage its pages.

This means that updating the number of pages and the content in the tabs is much easier, and we can resolve #706 .

This PR moves only the team list to ViewPager2. Future PRs will move other screens.
The initial work involved was twofold: First, pulling in ViewPager2 and doing the work of replacing the existing ViewPager, and second matching the styling of the new tab strip to the old one.

**Issues Reference:** 
#706 - Dynamic tab strips

**Test Plan:** 
Manually tested on a variety of devices. I also manually overrode our data loading to simulate loading additional/fewer teams to verify that the tabs and their contents updated appropriately. Seems to be working!

**Screenshots:**
Here's a screenshot to prove that the new tabs look the same as they did before!
teams_new

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -230,6 +230,8 @@ dependencies {
     implementation 'androidx.gridlayout:gridlayout:1.0.0'
     implementation 'androidx.multidex:multidex:2.0.1'
     implementation 'com.google.android.material:material:1.0.0'
+    implementation 'androidx.viewpager2:viewpager2:1.0.0-rc01'
+    implementation 'com.google.android.material:material:1.2.0-alpha01'
 
     // Play Services Libraries
     // See http://developer.android.com/google/play-services/setup.html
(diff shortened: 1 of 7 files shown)


phil-lopreiato:
Merged manually (I fixed the conflicts), thanks for doing this - I'm glad we'll be able to unblock more custom viewpagers :)

PR #953, opened 2020-06-22
[Do Not Merge] Sign In With Apple

arnavs-0:
Added Initial Sign In with Apple.

**Summary:** 
Added Initial Sign In with Apple. We need to return the user for the profile. Clean up Code, Add Apple Developers to Firebase. We need to add a warning for using user data.

**Test Plan:** 
Still under development. Need to add all features. 
**Screenshots:**
redirect
Home
redirect load

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -94,7 +94,7 @@ android {
 
     defaultConfig {
         applicationId "com.thebluealliance.androidclient"
-        minSdkVersion 16
+        minSdkVersion 17
         targetSdkVersion 29
         versionCode versionNum
         versionName version.toString()
@@ -215,6 +215,7 @@ repositories {
     maven {
         url "http://github.com/wada811/Android-Material-Design-Colors/raw/master/repository/"
     }
+    mavenCentral()
 }
 
 task checkstyle(type: Checkstyle) {
@@ -291,6 +292,7 @@ dependencies {
     implementation 'javax.annotation:javax.annotation-api:1.3.2'
     implementation "com.github.hotchemi:permissionsdispatcher:${permissionDispatcherVersion}"
     annotationProcessor "com.github.hotchemi:permissionsdispatcher-processor:${permissionDispatcherVersion}"
+    implementation 'com.willowtreeapps:signinwithapplebutton:0.3'
 
     // testing
     testImplementation 'androidx.test:core:1.2.0'
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -215,7 +215,6 @@ repositories {
     maven {
         url "http://github.com/wada811/Android-Material-Design-Colors/raw/master/repository/"
     }
-    mavenCentral()
 }
 
 task checkstyle(type: Checkstyle) {
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,26 +1,18 @@
-# Project-wide Gradle settings.
-
-# IDE (e.g. Android Studio) users:
-# Settings specified in this file will override any Gradle settings
-# configured through the IDE.
-
-# For more details on how to configure your build environment visit
+## For more details on how to configure your build environment visit
 # http://www.gradle.org/docs/current/userguide/build_environment.html
-
+#
 # Specifies the JVM arguments used for the daemon process.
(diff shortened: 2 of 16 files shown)


bherbst:
How many Android users do we expect regularly use Apple accounts?

android/build.gradle
```diff
@@ -94,7 +94,7 @@ android {
 
     defaultConfig {
         applicationId "com.thebluealliance.androidclient"
-        minSdkVersion 16
+        minSdkVersion 17
```
bherbst, line 97: @phil-lopreiato any stats on users on API level 16?

Honestly I'd be in favor of going to 21 or 23 soon, but just curious on whether there's any drive to keep it lowerl

android/build.gradle
```diff
@@ -215,6 +215,7 @@ repositories {
     maven {
         url "http://github.com/wada811/Android-Material-Design-Colors/raw/master/repository/"
     }
+    mavenCentral()
```
bherbst, line 218: Were you having issues resolving the new dependency without this? The top level build.gradle already defines `mavenCentral()` as a repo: https://github.com/the-blue-alliance/the-blue-alliance-android/blob/master/build.gradle#L41

gradle.properties
```diff
@@ -1,26 +1,18 @@
...
-
-# For more details on how to configure your build environment visit
+## For more details on how to configure your build environment visit
 # http://www.gradle.org/docs/current/userguide/build_environment.html
-
+#
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-# Default value: -Xmx10248m -XX:MaxPermSize=256m
+# Default value: -Xmx1024m -XX:MaxPermSize=256m
```
bherbst, line 6: Lol, that was definitely wrong before.

arnavs-0:
> How many Android users do we expect regularly use Apple accounts?

I don't exactly know how much we can expect. However, I came across a couple of articles explaining why we should have it as TBA has an iOS app. It gives people who have an android device to sign in with their own apple id. More and more people are getting apple ids from Android. I am not sure how it will impact us, however, I am seeing more Sign in With Apple on Android Apps.

android/build.gradle
arnavs-0: I did encounter dependency issues so I tried adding a few things and it worked when I added it worked. I'll try again to see if it was just Android Studio acting up.

ZachOrr:
@bherbst (it doesn’t seem like I can reply inline) re: how many Android users will use Sign In with Apple - the motivation behind this is Apple is requiring us on iOS to support Sign In with Apple, since we offer other social logins. Currently, we can’t push updates to the App Store (we’ve been blocked since April-ish) because we’re not in compliance with this new rule yet. Part of the challenge is we need to enable users to use Sign In with Apple on all of our platforms. So although signing in with an Apple ID will be rare on Android, we need it to ensure if users start their account on web with Sign In with Apple, they can pickup on Android

bherbst:
Ugh, forgot about that requirement for Apple sign-in. Big old 👎  to Apple.

android/build.gradle
arnavs-0: If it is a must we can keep Android 16. I just used a dependency for the Sign In with Apple button so I could test really quickly. If we needed I can make a button just thought this would be easier lol. The dependency needed a minimum of API 17

android/build.gradle
arnavs-0: Thanks for catching this. It worked, I guess Android Studio was acting up when I ran it.

android/build.gradle
phil-lopreiato: I've always just kept it where it was until we had a compelling enough reason to bump it. Bumping to 17 should have practically no effect

bherbst:
Closing out as stale - I don't foresee supporting Apple sign in on Android devices any time soon.

ZachOrr:
Closing this is likely fine. I believe Firebase will deduplicate accounts using the same email that sign in using different providers. So if your Apple ID is using your Gmail, you'll end up in the same spot.

PR #955, opened 2022-02-27
Migrate to FCM Cloud Messaging and dagger-android

phil-lopreiato:
Okay this became a whole huge thing.

Basically, since Android 12 is stricter about `PendingIntent` flags, the way we used to pull GCM tokens and register them broke, since the library used `PendingIntent` internally without the flags.

So, this will:
 - move registration to the new, shiny `WorkerManager` APIs, which replaces `JobIntentService`
 - migrates all the DI to `dagger-android`, because that was actually the easiest way to get the worker to be injected
 - use FCM APIs locally, not GCM


bherbst (approved):
Focused on the notification changes and they look good to me.

Love the hilt-ification! Mostly just skimmed those changes since it ended up touching all the things, but still very exciting.

🚀

phil-lopreiato:
So on the hilt stuff, it sort of spiraled. Old `InternService` stuff had to be migrated to `WorkManager`, which really can only be injected by hilt, which required redoing the whole app's DI...

PR #963, opened 2022-03-26
Fix geo intents on Android 11+

bherbst:
**Summary:** 
Trying to tap on an event's location on Android 11+ currently results in a toast that says "No app can handle that request".

This is because Android 11 and above require applications that query other packages to declare what they want to query in the manifest with `<query>` tags. See [package visibility](https://developer.android.com/training/package-visibility) for more info on this change.

The only reason we used this was to check if our links could be handled before launching them. Just launching the intent and catching any resulting `ActivityNotFoundException` has the same effect, so that's the route I went to fix.

**Test Plan:** 

Tested current behavior on Android 12L, and it works now!
I also changed the `geo:` intents to some garbage text to test the not found behavior, and confirmed the toast still displays as expected for unhandled intents.


phil-lopreiato:
Thanks for the fix!

PR #994, opened 2025-01-18
Migrate API Status model to new generated model

bherbst:
**Summary:** 
Migrate our first API models to the new openapi-gen models created in #992 - `ApiStatus`.


bherbst:
Ope, that didn't work the way I expected

PR #1027, opened 2026-02-12
Refactor events screen organization

bherbst:
I was going to play around with a double elimination UI, but right now all the event detail screen is in one large file - I've split that out so each tab is in its own isolated file to clean up the organization. No code has changed within the tabs.


gregmarra (approved):
Clean refactor — all tab bodies are identical to the originals, just moved to focused files under a sensible package structure. Two tiny nits: `LoadingBox.kt` is missing a trailing newline, and it uses 2-space indentation while `EmptyBox.kt` uses 4-space.

PR #1028, opened 2026-02-12
Add monochrome icon

bherbst:
Android 13 introduced support for monochrome app icons, where the icon takes on coloring from the user's system-wide theme.

This PR adds support for monochrome icons by simply re-using our foreground icon vector. 

|On a device with icon theming|Device without theming|
|-|-|
|||


gregmarra (approved):
Looks great — clean, minimal change. Reusing the foreground drawable for monochrome is the standard approach and the screenshots look good.

PR #1059, opened 2026-02-14
Add automated release builds to GitHub Releases

dracco1993:
**Summary:**
Automatically post a pre-release APK to GitHub Releases when new code is pushed to `main.

The only real thought/question I have is if there's some better way to do versioning; I'm not at all tied to how it's done now, and would totally be open to other ideas!

**Issues Reference:**
N/A

**Test Plan:**
Tested on my fork:
  - https://github.com/dracco1993/the-blue-alliance-android/releases/tag/build-1

**Screenshots:**
N/A

### DIFF
--- a/copilot-setup-steps.yml
+++ b/copilot-setup-steps.yml
@@ -1,0 +1,65 @@
+name: Setup Steps for The Blue Alliance Android
+
+steps:
+  - name: Fetch full git history and tags
+    run: |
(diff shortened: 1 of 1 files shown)


phil-lopreiato:
I wonder if we can use github's "deployments" feature for these?

Or maybe we can make this a cron that publishes nightly builds?

PR #1068, opened 2026-02-18
Add event and team app shortcuts

bherbst:
Adds dynamic app shortcuts for signed-in users' favorite events and teams.

There isn't a lot of smarts behind selecting the ~4 shortcuts that will appear for a user right now, but at least gives a starting point for the future. We could also experiment with publishing a dynamic shortcut whenever someone _views_ any team or event instead of favorites.

I was really struggling to get MyTBA working for me locally, but I mocked out the responses in what I think is a realistic manner for testing. If someone else has a working local MyTBA setup would love more testing though 🙏 

Screenshot of what these can look like on the homescreen:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ The Blue Alliance is built by volunteers. We'd love your help!
 - **Teams** -- Browse teams, view team details with year-by-year event participation and media
 - **Districts** -- Browse district listings and rankings
 - **Search** -- Find teams and events across all of TBA
-- **myTBA** -- Sign in with Google to save favorite teams and events, and subscribe to push notifications for match scores, upcoming matches, schedule changes, and more
+- **myTBA** -- Sign in with Google to save favorite teams and events, and set up push notifications for match scores, upcoming matches, schedule changes, and more
 - **Deep linking** -- Open thebluealliance.com links directly in the app
 - **Push notifications** -- Receive alerts for match scores, upcoming matches, event updates, and more via Firebase Cloud Messaging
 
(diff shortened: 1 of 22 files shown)


gregmarra:
For icons, I prefer the "people" icon we're using in the tab bars for teams - I like the idea that teams are the people that build the robot, not the robot. BUT with the Avatar code landed, maybe we can use the Avatar image to get the team's own identity in there?

For events, we're really tight on space - should we drop year? Should we use event codes? Hmm.

bherbst:
>For icons, I prefer the "people" icon we're using in the tab bars for teams - I like the idea that teams are the people that build the robot, not the robot. BUT with the Avatar code landed, maybe we can use the Avatar image to get the team's own identity in there?

You read my mind - I was waiting for that to merge then I was going to swap over to avatars. Will get that going!

> For events, we're really tight on space - should we drop year? Should we use event codes? Hmm.
Yeah, I don't love that right now either. I considered event codes and would definitely be open to that! Maybe not the most user-friendly display, but would fit better. Dropping the year _might_ be okay because folks _probably_ aren't keeping shortcuts around a year later? But wouldn't be very clear what the shortcut goes to if someone does have it hanging around for a while.

bherbst:
Still working on getting a local TBA server working correctly, but did push some changes to use team avatars for those shortcuts.

I don't _love_ that they are square pegs in round holes, but I also don't think cropping would be a great look for a lot of avatars either

image

gregmarra (approved):

PR #1094, opened 2026-02-24
Give each screen its own scaffold

### LINKED ISSUE #1081: Screen titles don't display properly during predictive back gestures
See recording below - when animating between screens (particularly noticeable with predictive back), the screen title doesn't update until the animation is complete. This results in the title kind of jumping in or out during the animation.

The "right" fix here is likely to move towards each screen maintaining its own `Scaffold` with its own app bar instead of updating state on a shared bar. This also would give each screen much more flexibility in what their app bar looks like the future.

https://github.com/user-attachments/assets/a8bcaf31-3ce4-4442-a9be-615435074efe

bherbst:
Replace the shared top-level `Scaffold` and `TopAppBar` with a per-screen `Scaffold` and `TopAppBar`. This reduces the complexity in managing a shared global state and gives individual screens more breathing room to customize what they show in the top app bar. The bottom bar remains a shared top-level component.

To maintain reasonable transition animations between screens, this PR introduces a new transition animation between tabs that does a simple cross fade.

This resolves #1081 as well.

Work was primarily done with Copilot.

https://github.com/user-attachments/assets/68bc26ac-b0df-464e-864c-70d252845c41

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -25,7 +25,7 @@ datastore = "1.2.0"
 coil = "2.7.0"
 aboutlibraries = "14.0.0-b02"
 
-firebase-bom = "34.9.0"
+firebase-bom = "34.10.0"
 google-services = "4.4.4"
 firebase-crashlytics-plugin = "3.0.6"
 credentials = "1.5.0"


gregmarra (commented):
Nice work! The per-screen scaffold is the right call — the old shared TopAppBar conditional logic was getting unwieldy. The `MutableSharedFlow<Unit>` for reselection, `TBABottomBar` extraction, and `Transitions.kt` are all clean. Every screen now has pretty similar Scaffold+TopAppBar+back+search boilerplate — could be worth extracting a shared `TBATopAppBar` as a follow-up to keep things DRY, but not blocking.

app/src/main/kotlin/com/thebluealliance/android/navigation/TBANavigation.kt
```diff
@@ -24,177 +33,243 @@ import com.thebluealliance.android.ui.teamevent.TeamEventDetailViewModel
...
+import androidx.compose.foundation.lazy.LazyListState
+import androidx.compose.foundation.lazy.rememberLazyListState
+import androidx.compose.runtime.mutableStateMapOf
+import androidx.compose.runtime.rememberCoroutineScope
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.emptyFlow
+import kotlinx.coroutines.flow.filter
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.launch
```
gregmarra: Looks like there are a bunch of unused imports in this file — `coroutineScope`, `mutableStateMapOf`, `mutableIntStateOf`, `windowInsetsPadding`, `WindowInsets`, `WindowInsetsSides`, `only`, `safeDrawing`, `LazyListState`, `rememberLazyListState`, `filter`, `map`. Same in a couple other files (`DistrictsScreen.kt` and `TeamsScreen.kt` have unused `viewModel` imports). An "Optimize Imports" pass across the modified files would clean these up.

PR #1105, opened 2026-02-27
Replace "Happening Now" with "This Week" section

gregmarra:
## Summary
- Replaces the narrow "Happening now" section (only events whose dates include today) with a broader "This Week" section that shows all events for the current competition week
- During regular season, uses the TBA API `week` field to group events ("This Week — Week N"); during offseason/preseason, falls back to calendar-week overlap ("This Week")
- Prevents prematurely showing the next competition week during preseason by requiring at least one weeked event to have ended before using the upcoming-week lookahead

## How it works
The algorithm in `computeThisWeekEvents()` detects the current competition week by checking:
1. Events with a `week` value happening today
2. Upcoming weeked events within 7 days (only if the season has started)
3. Recently-ended weeked events within 3 days
4. Fallback: calendar-week overlap (Monday–Sunday)

Championship events (types 3/4, `week=null`) are included via date overlap when they coincide with a regular-season week's date range. The section is hidden when viewing a non-current year or when no events match.

## Screenshot
image

## Test plan
- [x] Unit tests for `computeThisWeekEvents`: regular season, Sunday gap, preseason guard, offseason fallback, championship inclusion, non-current year, empty events
- [x] `./gradlew :app:assembleDebug` builds clean
- [x] Verified on emulator: during preseason (Feb 27), shows "This Week" with preseason events, not Week 1

### DIFF
--- a/app/src/main/kotlin/com/thebluealliance/android/ui/events/EventsUiState.kt
+++ b/app/src/main/kotlin/com/thebluealliance/android/ui/events/EventsUiState.kt
@@ -54,13 +54,13 @@ fun computeThisWeekEvents(
         }
 
         events = (weekEvents + championshipEvents).distinctBy { it.key }
-        label = "Happening This Week \u2014 Week ${currentWeek + 1}"
+        label = "Upcoming This Week \u2014 Week ${currentWeek + 1}"
     } else {
         // Offseason fallback: calendar week overlap (Monday-Sunday)
         val monday = today.with(DayOfWeek.MONDAY)
         val sunday = monday.plusDays(6)
         events = allEvents.filter { datesOverlap(it, monday, sunday) }
-        label = "Happening This Week"
+        label = "Upcoming This Week"
     }
 
     // Remove events that have already ended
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/com/thebluealliance/android/ui/events/EventsUiState.kt
```diff
@@ -12,3 +15,114 @@ sealed interface EventsUiState {
...
+
+    // 1. If any event with a week is happening today → use that week
+    val happeningToday = weekedEvents.filter { event ->
+        val start = parseDate(event.startDate)
+        val end = parseDate(event.endDate)
+        start != null && end != null && !today.isBefore(start) && !today.isAfter(end)
+    }
+    if (happeningToday.isNotEmpty()) {
+        return happeningToday.first().week
+    }
```
bherbst: I'm not sure how the Event models' weeks are calculated on the backend - 

If there's an event that starts the first Saturday of competition (Week 1?)  and ends the next day, what week is "this week"  on Sunday - 1 (because a week 1 event is still running) or 2 (because most of the events coming up are week 2)?

PR #1106, opened 2026-02-27
Search: Allow single-character search, scroll to top

bherbst:
Previously it was difficult to search for single-digit teams because we didn't start actually searching until you had 2 characters in the search box.

This also fixes a bug where if our events query came back before the teams did, the list would be scrolled to events instead of the top of the page. 

Realistically, whenever we get new search results we should just go back to the top of the page because the results could be totally different.


gregmarra (approved):
Looks good! Minor nit: `query.length  == 0` has a double space and could be `query.isEmpty()`, but not blocking.

PR #1135, opened 2026-03-02
[don't merge] Add match tracking with live updates notification

gregmarra:
Adds a match tracking foreground service — pick a team, get a persistent notification that shows what match is up next, what's playing now, and scores as they come in. Designed for FRC event attendees who want to follow a specific team throughout a competition day.

## Architecture

The tracking system lives in the `tracking/` package with four files:

- **`MatchTrackingService`** — foreground service that owns the notification lifecycle. Runs three concurrent coroutine jobs: an `observeJob` that watches Room for match/event data changes, a `tickerJob` that re-evaluates the notification against wall-clock time (so "Next: Q42 est. 2:30 PM" transitions to "NOW PLAYING Q42" promptly), and a `refreshJob` that polls the TBA API as a safety net for missed FCM pushes. All three use adaptive intervals — faster when a match is imminent, slower during gaps.
- **`MatchTrackingState`** — pure functions that compute `TrackedTeamState` from a match list. Determines next/current/last match, qual record, and auto-dismiss timing. No Android dependencies, fully unit-testable.
- **`MatchTrackingNotificationBuilder`** — builds the notification from state. Handles promoted ongoing (API 35+), status bar chip text (API 36+), and a dormant variant for overnight mode.
- **`MatchTrackingManager`** — high-level coordinator injected into ViewModels. Finds the team's current event, refreshes data, starts/stops the service.

### Adaptive polling and ticker

Both the API poll interval and the ticker interval adapt based on how soon the team plays next:

| Scenario | API poll | Ticker |
|---|---|---|
| Match imminent (< 15 min) or in progress | 5 min | 60s |
| Match within the hour | 15 min | 5 min |
| No match soon / overnight | 60 min | 5 min |

The ticker also wakes 1 minute before a match's estimated start so the notification transitions to "now" promptly.

### Dormant mode for multi-day events

FRC events run Friday–Saturday (sometimes Sunday). Without dormant mode, the 2h auto-dismiss after the last Friday match would kill the service — users would have to re-enable tracking Saturday morning.

When auto-dismiss fires but the event's `endDate` hasn't passed, the service enters dormant mode instead of stopping:
- Minimal low-priority notification ("175 — Tracking paused")
- Ticker cancelled (no UI transitions to track overnight)
- API polling slowed to 60 min (just watching for next-day schedule to appear)
- When a poll discovers the team has an upcoming match (or is currently playing), exits dormant and resumes full tracking
- On the morning after the last event day, the dormant check stops the service entirely

### UI integration

Tracking is started from the team detail screen's notification preferences sheet or from favorited teams in the myTBA tab. `MatchTrackingService.activeTeamKey` (a `StateFlow`) lets the UI reflect tracking state — showing "Stop" vs "Track" buttons, confirming before switching teams, etc.

## Test plan

- [x] Unit tests for `computeTrackedTeamState`, adaptive polling/ticker delays, `shouldEnterDormant`
- [ ] Manual test on `2026test` — start tracking, verify notification updates, auto-dismiss, dormant entry/exit
- [ ] Verify notification renders correctly on API 35+ (promoted ongoing) and API 36+ (chip text)

### DIFF
--- a/app/src/main/kotlin/com/thebluealliance/android/tracking/MatchTrackingNotificationBuilder.kt
+++ b/app/src/main/kotlin/com/thebluealliance/android/tracking/MatchTrackingNotificationBuilder.kt
@@ -18,6 +18,16 @@ object MatchTrackingNotificationBuilder {
 
     const val NOTIFICATION_ID = 7175 // "TBA Team 175" :)
 
+    private fun stopPendingIntent(context: Context): PendingIntent {
+        val stopIntent = Intent(context, MatchTrackingService::class.java).apply {
+            action = MatchTrackingService.ACTION_STOP
+        }
+        return PendingIntent.getService(
+            context, 0, stopIntent,
+            PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE,
+        )
+    }
+
     fun build(
         context: Context,
         state: TrackedTeamState,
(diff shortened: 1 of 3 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -55,6 +57,15 @@
             </intent-filter>
         </service>
 
+        <service
+            android:name=".tracking.MatchTrackingService"
+            android:exported="false"
+            android:foregroundServiceType="specialUse">
+            <property
+                android:name="android.app.PROPERTY_SPECIAL_USE_FGS_SUBTYPE"
+                android:value="Live match tracking for a team at an FRC competition event" />
```
bherbst: Worth noting - this [will trigger additional scrutiny during Google Play review](https://developer.android.com/develop/background-work/services/fgs/service-types#special-use), and if Google disagrees with the use case they may reject the update.

PR #1138, opened 2026-03-02
Generate startup and baseline profiles during release

### LINKED ISSUE #1070: Set up startup profiles
We will likely get a decent performance boost by [configuring a startup profile](https://developer.android.com/topic/performance/startupprofiles/dex-layout-optimizations).

Ideally the startup profile generation would be part of the release CI process

bherbst:
This PR adds a new `baselineprofile` module that generates [baseline](https://developer.android.com/topic/performance/baselineprofiles/overview) & [startup](https://developer.android.com/topic/performance/baselineprofiles/overview#startup-profiles) profiles for the application. This should give us a bit of a performance improvement, particularly for early installs before cloud profiles get generated.

The startup profile benchmark test simply launches the app.
The baseline profile test will launch the app, navigate to the teams tab, navigate to the districts tab, and perform a simple search query. The intention is to cover a small subset of common functionality to build a reasonable baseline profile. We could add more flows later, but this at least gives us a start! [Here are Google's docs on what to include](https://developer.android.com/topic/performance/baselineprofiles/overview#what-to-include) for reference.

Our `release.sh` script (@gregmarra let me know if you have a different release process, but I'm hoping you just run that script) will first run these tests to generate the profiles, then the step to build a release app bundle will include the profiles automatically.

The tests are currently configured to run on a [Gradle-Managed Device](https://developer.android.com/studio/test/managed-devices) to help provide test device consistency, and we are using an ["automated test device" image](https://developer.android.com/studio/test/managed-devices#gmd-atd) that turns off a few unneeded features to help keep tests running as quickly as possible.

Resolves #1070

### DIFF
--- a/baselineprofile/src/main/java/com/thebluealliance/android/baselineprofile/BaselineProfileGenerator.kt
+++ b/baselineprofile/src/main/java/com/thebluealliance/android/baselineprofile/BaselineProfileGenerator.kt
@@ -1,4 +1,4 @@
-package com.thebluealliance.android.baseline
+package com.thebluealliance.android.baselineprofile
 
 import androidx.benchmark.macro.junit4.BaselineProfileRule
 import androidx.test.ext.junit.runners.AndroidJUnit4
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/com/thebluealliance/android/ui/events/EventsScreen.kt
```diff
@@ -74,6 +75,13 @@ fun EventsScreen(
         }
     }
 
+    // Report fully drawn once we have data
+    // https://developer.android.com/topic/performance/vitals/launch-time#startup-accuracy
+    ReportDrawnWhen {
+        uiState !is EventsUiState.Loading
+    }
+
+
```
gregmarra: extra newline

gregmarra (commented):
Looks great overall — two small things remaining, then I'm happy to approve.

gregmarra (approved):
All feedback addressed — package names match, extra newline cleaned up. lgtm!

PR #1188, opened 2026-03-08
Relocate red/blue teams list on match detail

KyleJ924:
**Summary:** 

- Minor adjustment match participant display to make better use of space
- Added bonus RP indicators

**Previous:**
image

**Updated:**
image


phil-lopreiato (approved):
Nice! I like the RP bonus dots being here too :)

PR #1194, opened 2026-03-08
Reorder events on events screen

KyleJ924:
**Summary:** 

- Reorders event listing
- Added subsection headers for Regionals and individual districts with districts sorted alphabetically
- Events in each subsection sorted alphabetically rather than by start date
- Subsection header color follows color of bottom navigation bar when switching light/dark mode

**Screenshots:**
image

### DIFF
--- a/app/src/main/kotlin/com/thebluealliance/android/data/mappers/Mappers.kt
+++ b/app/src/main/kotlin/com/thebluealliance/android/data/mappers/Mappers.kt
@@ -274,6 +274,17 @@ fun DistrictRankingEntity.toDomain() = DistrictRanking(
     rank = rank,
     pointTotal = pointTotal,
     rookieBonus = rookieBonus,
+    eventPoints = try {
+        json.decodeFromString<List<DistrictEventPointsDto>>(eventPoints).map {
+            DistrictEventPoints(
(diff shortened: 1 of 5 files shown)


gregmarra (commented):
Nice work on the district subsection headers! A couple of things to address:

## `buildSubSections` treats all non-district events as "Regional Events"

In `EventsUiState.kt`, `buildSubSections` splits events into `district == null` → "Regional Events" and everything else by district key. But preseason, offseason, and championship events also have `district == null`, so they'd show up under a "Regional Events" header — which isn't accurate.

You might want to filter by event type as well — regionals are type 0, so something like:
```kotlin
val regionals = events.filter { it.district == null && it.type == 0 }
val other = events.filter { it.district == null && it.type != 0 }
```
...and either give `other` its own subsection label (or no subsection header at all), or just pass non-district, non-regional events through without a subsection.

## Sticky header index calculation could use a test

The index math changed from `1 + section.events.size` to `1 + section.subSections.sumOf { 1 + it.events.size }` — it would be great to have a test that builds sections with subsections and verifies the header indices land correctly. This is the kind of thing that's easy to get off-by-one on and hard to catch visually.

PR #1201, opened 2026-03-09
Simplify widget ProGuard rules to single wildcard

gregmarra:
## Summary
- Replaces per-class `-keep` rules for each `ActionCallback` subclass with a single wildcard rule that covers all current and future implementations
- Same protection (class name + no-arg constructor preserved by R8), less maintenance

## Test plan
- [x] Built release APK with `tba-keys` signing
- [x] Installed on emulator, verified widget config dialog opens
- [x] Previously confirmed on Pixel 9 Pro that the verbose version of this rule fixes widget tap actions — this is a pure simplification of the same rule


bherbst (approved):
Surprised glance doesn't ship this rule itself, but looks good to me

PR #1357, opened 2026-05-29
perf: add Room indices on hot filter columns

gregmarra:
## Summary
Several DAO `observe*`/`delete*` queries filter on non-primary-key columns, forcing full table scans. Adds `@Index` on the genuinely uncovered hot columns and bumps the DB version to 16:

- `matches.eventKey` (PK is `key`)
- `event_teams.teamKey` (PK leads with `eventKey`)
- `team_event_status.eventKey` (PK leads with `teamKey`)
- `events.year`, `events.district` (PK is `key`)
- `districts.year` (PK is `key`)

Composite-PK tables already index their leading column, so only filters not covered by an existing index are added. Destructive migration is already configured, so the cache is simply re-fetched on upgrade — no migration code needed.

## Test plan
- [x] `./gradlew :app:assembleDebug` passes (Room schema validation)
- [x] Verified on emulator (2026-05-29): DB migrated to `user_version=16` and all 6 indices are present (`index_districts_year`, `index_event_teams_teamKey`, `index_events_year`, `index_events_district`, `index_matches_eventKey`, `index_team_event_status_eventKey`).

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -39,7 +39,7 @@ core-splashscreen = "1.2.0"
 
 junit = "6.1.0"
 turbine = "1.2.1"
-mockk = "1.14.9"
+mockk = "1.14.11"
 play-publisher = "4.0.0"
 lifecycle-process = "2.10.0"
 ktlint-gradle = "14.2.0"
--- a/app/src/main/kotlin/com/thebluealliance/android/data/local/TBADatabase.kt
+++ b/app/src/main/kotlin/com/thebluealliance/android/data/local/TBADatabase.kt
@@ -60,7 +60,7 @@ import com.thebluealliance.android.data.local.entity.TeamEventStatusEntity
         EventRankingSortOrderEntity::class,
         TeamEventStatusEntity::class,
     ],
-    version = 15,
+    version = 16,
     exportSchema = false,
 )
 abstract class TBADatabase : RoomDatabase() {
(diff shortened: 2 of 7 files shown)


bherbst (approved):

gradle/libs.versions.toml
```diff
@@ -39,7 +39,7 @@ core-splashscreen = "1.2.0"
 
 junit = "6.1.0"
 turbine = "1.2.1"
-mockk = "1.14.9"
+mockk = "1.14.11"
```
bherbst, line 42: You may have already noticed this, but all your PRs bump this so you're going to get a conflict on pretty much all of them

gregmarra: Yeah, that bump is on all 8 — it's a workaround for the `NewerVersionAvailable` lint error that started failing on `main` once mockk 1.14.11 was published (lint has `abortOnError`, so every PR's `android-lint` job went red regardless of its actual changes). Each branch needed it just to get CI green.

Since they're all the identical `1.14.9` → `1.14.11` edit, a 3-way merge treats them as the same change and auto-resolves — so merging them in sequence shouldn't produce a real conflict. But if you'd rather keep it clean, I'm happy to pull the bump into a single dedicated PR, merge that first, and drop it from the other 7. Your call on which you prefer.

bherbst: Nah totally fine - I was assuming Git would probably figure it out okay.

gregmarra:
Thanks! Leaving the bump on each branch then — git 3-way-merges the identical edits cleanly as they go in.

PR #1360, opened 2026-05-29
perf: only schedule Wear periodic refresh when a team is tracked

gregmarra:
## Summary
- `WearApplication.onCreate` unconditionally enqueued the 6-hour periodic complication refresh, so the worker woke every 6h even with no team configured and no complication present — needless wakeups/battery.
- Gates the enqueue on a non-blank tracked team, and (re)schedules the periodic refresh from `TeamTrackerActivity` and `TeamTrackingComplicationConfigActivity` when a team is set, so it still starts as soon as a team is chosen.
- Per Bryan's review: the worker now also cancels the periodic refresh in its `doWork()` `else` branch when it runs and finds no tracked team, so it stops the 6h wakeups (re-enqueued when a team is set again).
- `enqueueUniquePeriodicWork` uses `KEEP`, so the repeated calls are idempotent.

## Test plan
- [x] `./gradlew :wear:assembleDebug` passes
- [x] Verified on wear emulator (2026-05-29) via the WorkManager diagnostics broadcast: setting a team takes the periodic job count 0→1; relaunching with a blank team shows `complication_refresh` → **CANCELLED** and `complication_immediate_refresh` → **SUCCEEDED**, confirming the worker cancels the periodic work when no team is tracked.

### DIFF
--- a/wear/src/main/kotlin/com/thebluealliance/android/wear/WearApplication.kt
+++ b/wear/src/main/kotlin/com/thebluealliance/android/wear/WearApplication.kt
@@ -4,6 +4,7 @@ import android.app.Application
 import androidx.hilt.work.HiltWorkerFactory
 import androidx.work.Configuration
 import com.thebluealliance.android.wear.data.ApiKeyProvider
+import com.thebluealliance.android.wear.tracker.TeamTrackerPreferences
 import com.thebluealliance.android.wear.worker.TeamTrackingComplicationWorker
 import dagger.hilt.android.HiltAndroidApp
 import javax.inject.Inject
@@ -26,6 +27,11 @@ class WearApplication :
     override fun onCreate() {
         super.onCreate()
         apiKeyProvider.init()
-        TeamTrackingComplicationWorker.enqueuePeriodicRefresh(this)
+        // Only schedule the periodic refresh when a team is actually tracked;
+        // otherwise the worker wakes every 6h to do nothing. Setting a team
+        // schedules it from the tracker/config activities.
+        if (TeamTrackerPreferences(this).teamNumber.isNotBlank()) {
+            TeamTrackingComplicationWorker.enqueuePeriodicRefresh(this)
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


wear/src/main/kotlin/com/thebluealliance/android/wear/WearApplication.kt
```diff
@@ -26,6 +27,11 @@ class WearApplication :
     override fun onCreate() {
         super.onCreate()
         apiKeyProvider.init()
-        TeamTrackingComplicationWorker.enqueuePeriodicRefresh(this)
+        // Only schedule the periodic refresh when a team is actually tracked;
+        // otherwise the worker wakes every 6h to do nothing. Setting a team
+        // schedules it from the tracker/config activities.
+        if (TeamTrackerPreferences(this).teamNumber.isNotBlank()) {
+            TeamTrackingComplicationWorker.enqueuePeriodicRefresh(this)
```
bherbst, line 34: Should the worker also be cancelling the periodic work if it runs and there is no longer any team being tracked?

bherbst (approved):

wear/src/main/kotlin/com/thebluealliance/android/wear/WearApplication.kt
gregmarra: Good call — done. Added an `else` branch in `doWork()` so when the worker runs and finds no tracked team it calls `cancelPeriodicRefresh()`, stopping the 6h wakeups. It gets re-enqueued when a team is set again (config activity / tracker / app onCreate).

gregmarra:
@bherbst heads up — I pushed the cancel-on-no-team change you suggested *after* your approval, so your green check is on the prior commit. Mind taking another quick look before this merges? Verified on the wear emulator via the WorkManager diagnostics broadcast: relaunching with a blank team shows `complication_refresh` → CANCELLED and `complication_immediate_refresh` → SUCCEEDED.

bherbst (approved):

PR #1361, opened 2026-05-30
RFC: Fix district/regional advancement points data-loss race

gregmarra:
**RFC + implementation.** The write-up lives here in the description (no committed `.md`); the diff is the actual fix, built and unit-tested.

## Problem

District points and regional points are two distinct TBA datasets (`getEventDistrictPoints` / `getEventRegionalPoints`, which share `EventDistrictPointsResponseDto`), but they were stored in **one Room table with no discriminator** — `event_district_points`, PK `(eventKey, teamKey)`.

In `EventRepository` both observe methods read `observeByEvent(eventKey)` and both refresh methods ran `deleteByEvent(eventKey)` then insert. So:

- `EventDetailViewModel` *observes* the correct dataset conditionally (`event.district != null`) but *refreshes* **both endpoints unconditionally, in parallel**.
- For a district event, `getEventRegionalPoints` returns an empty `200` → `deleteByEvent` wiped the real district rows and inserted nothing → the Points tab blanked non-deterministically depending on which parallel refresh landed last. Errors were swallowed by `catch (_: Exception) {}`.
- Secondary: `TeamEventDetailViewModel` mapped the same physical rows into two "different" maps that were always identical.

## Fix

Give the shared table a discriminator so both datasets coexist, and name it honestly for what it holds.

- **Discriminator:** add a `source: PointsSource` column to the primary key (`PointsSource` is a Kotlin enum, `DISTRICT` / `REGIONAL`; Room 2.8.4 persists it natively via the built-in enum converter — no `@TypeConverter`). Each refresh only clears its own source, so an empty regional response can no longer wipe district rows.
- **Honest naming:** the table held district *and* regional points all along, so rename `event_district_points` → `event_advancement_points`, `EventDistrictPointsEntity` → `EventAdvancementPointsEntity`, `EventDistrictPointsDao` → `EventAdvancementPointsDao`. PK is `(eventKey, teamKey, source)`; `observeByEvent` / `deleteByEvent` take a `source` arg.
- `TBADatabase` version `16 → 18`. The DB uses `fallbackToDestructiveMigration(dropAllTables = true)` over cached API data, so no hand-written migration and no user-visible data loss.

Public `EventRepository` methods (`observe`/`refresh` × `District`/`Regional`) and the TBA DTOs keep their district/regional names, since those map to two distinct API endpoints.

The ViewModels need **no change**: each endpoint now owns its rows, so `EventDetailViewModel`'s dual-refresh is harmless and `TeamEventDetailViewModel`'s two maps become genuinely distinct.

## Tests

Two regression tests in `EventRepositoryTest` (mockk):
- `refreshEventRegionalPoints` deletes only `PointsSource.REGIONAL` and **never** `PointsSource.DISTRICT` (the data-loss guard).
- `refreshEventDistrictPoints` tags inserted rows with `source = PointsSource.DISTRICT`.

`./gradlew :app:ktlintCheck :app:testDebugUnitTest` → BUILD SUCCESSFUL.

## Risk

Low — destructive cache migration only; no user data. Touches entity, DAO, mapper, repository (4 methods), DB version, and tests. No ViewModel/UI changes.


gregmarra:
This is a great fix for the data integrity issue where district and regional points were overwriting each other. Bumping the database version with destructive migration is also completely safe here since it acts as a client-side cache.

One recommendation to improve type safety:

### Use an Enum Class for the `source` Column
Currently, the codebase relies on raw string literals (`"district"` and `"regional"`) declared privately in `EventRepository.kt` and duplicated inside tests in `EventRepositoryTest.kt`. 

Since Room supports Kotlin Enum classes natively, we can replace the raw strings with a dedicated enum to enforce compile-time safety and prevent typos:

1. **Define the Enum class**:
   ```kotlin
   package com.thebluealliance.android.data.local.entity

   enum class PointsSource {
       DISTRICT,
       REGIONAL
   }
   ```

2. **Use it in `EventDistrictPointsEntity.kt`**:
   ```kotlin
   @Entity(tableName = "event_district_points", primaryKeys = ["eventKey", "teamKey", "source"])
   data class EventDistrictPointsEntity(
       val eventKey: String,
       val teamKey: String,
       val source: PointsSource,
       // ...
   )
   ```

This avoids string duplication in the repository and tests, making it clear what the allowed sources are.

gregmarra:
Agreed this is a real type-safety improvement, and you're right that Room handles it natively — we're on Room 2.8.4, which ships a default enum converter (2.3+), so no manual `@TypeConverter` would be needed.

I'd like to keep this PR scoped to just the data-loss race fix, though. The enum is an orthogonal schema/type-safety change that touches the entity, `EventRepository`, and the tests, so it reads more cleanly as its own focused cleanup PR (same reasoning that led to splitting #1362 into separate PRs). Since this PR already bumps the DB version with a destructive migration, folding the column-type change into a follow-up costs nothing extra.

I'll leave the string constants here and open a follow-up to introduce `PointsSource`. Declining for this PR only.

gregmarra:
Update: reversing my call above — we folded this into this PR rather than a follow-up.

Two things changed my mind:

1. The `"district"` / `"regional"` strings were **net-new in this PR** — the `POINTS_SOURCE_*` constants and their duplication in the tests didn't exist before. So "keep it scoped" really meant shipping new stringly-typed code and immediately scheduling its own cleanup. Cheaper to do it right once than to merge it and then un-merge it.
2. Looking harder at the storage itself, the single table was already misnamed: it holds district **and** regional rows, so `event_district_points` / `EventDistrictPointsEntity` were lying about their contents.

So this PR now:
- types the column as `enum class PointsSource { DISTRICT, REGIONAL }` (Room 2.8.4 persists it natively, no `@TypeConverter`), dropping the string constants and their test duplication; and
- renames the storage to match what it stores: `event_advancement_points` / `EventAdvancementPointsEntity` / `EventAdvancementPointsDao`, keeping `PointsSource` as the discriminator rather than splitting into two byte-identical tables (they share one DTO + one domain model and are mutually exclusive per event).

DB version bump is now `16 → 18` (the table rename changes Room's schema identity hash). Public repository methods and the TBA DTOs keep their district/regional names — those map to two distinct endpoints. Build + unit tests green.

PR #1363, opened 2026-05-30
RFC: Shared UiState<T> + screen scaffold for the 10 screens

gregmarra:
## RFC: Shared `UiState<T>` + screen scaffold

**Status:** Implemented across all 10 screens in this PR — the primitives plus every migration.

### Motivation
Every screen/VM re-implements loading / empty / error / refresh by hand, and they don't even agree on how:
- The **`_isRefreshing` triplet** (`MutableStateFlow` + `asStateFlow()` + the `try/finally` toggle) is copy-pasted into all 10 ViewModels.
- **Two incompatible `UiState` conventions** coexist — a sealed `Loading/Success` in some screens, a nullable data-class in others; only `EventsUiState` bothers with an `Error` case.
- `common/LoadingBox.kt` and `common/EmptyBox.kt` already exist but are bypassed by ~8 hand-rolled loading/empty boxes.
- **34 empty `catch (_: Exception) {}`** blocks silently swallow every network error.

### What this PR lands

Four reusable primitives in `ui/common/`:

**1. `UiState<out T>`** — one canonical, covariant sealed interface:
```kotlin
sealed interface UiState<out T> {
    data object Loading : UiState<Nothing>
    data class Success<T>(val data: T) : UiState<T>
    data object Empty : UiState<Nothing>
    data class Error(val message: String? = null) : UiState<Nothing>
}
```

**2. `StateContent(state, isRefreshing, onRefresh) { data -> }`** — wraps the body in `PullToRefreshBox` and renders the right surface for each state, reusing the existing `LoadingBox`/`EmptyBox`/`ErrorBox`. Callers supply only the success content.

**3. `ErrorBox(message, onRetry)`** — the missing sibling of `LoadingBox`/`EmptyBox`; centered message + optional Retry button.

**4. `RefreshableViewModel.refreshing(vararg tasks)`** — base class that owns the `isRefreshing` flag and runs refresh tasks concurrently under one `runCatching` that **logs instead of swallowing**.

### Migration — all 10 screens

The screens split into two tiers by shape:

**Single-content screens → full `UiState<T>` + `StateContent`** (Teams, RegionalAdvancement, Events, plus Districts as the original worked example). Per-screen sealed `UiState` types are deleted (`DistrictsUiState`, `TeamsUiState`, `RegionalAdvancementUiState`; `EventsUiState` collapses to a plain `EventsData` holder). The hand-rolled `Box` / `CircularProgressIndicator` / `PullToRefreshBox` plumbing collapses into a single `StateContent(...) { data -> ... }` call. RegionalAdvancement now actively **surfaces fetch errors** with Retry — the old silent `catch (_: Exception) {}` becomes `UiState.Error`. Teams, Events, and Districts stay behavior-preserving over their Room-cache-backed flows (empty maps to `Loading`/`Empty`, no error-flash).

**Detail screens → `RefreshableViewModel` only** (DistrictDetail, TeamDetail, EventDetail, MatchDetail, TeamEventDetail, plus MyTBA). These aggregate many fields into one `*DetailUiState`, so a single `UiState<T>` wrapper doesn't fit; instead they drop their duplicated `_isRefreshing` triplet and swallowed `catch {}` blocks for the shared `refreshing()` helper, which runs tasks concurrently with per-task failure isolation and **logs** failures rather than discarding them. Their aggregate state flows are unchanged.

**Search is left as-is** — it has no async load/refresh surface to share.

Net across the PR: **−582 lines** (1111 deleted, 529 added) over 22 files.

### Testing
- Every migrated surface emulator-verified without regressions: Teams list, Events sections, Districts list, RegionalAdvancement (including the new Error + Retry path on a year with no published data), and the five detail screens (District / Team / Event / Match / TeamEvent) + MyTBA all render and pull-to-refresh.


gregmarra:
Here are a few suggestions to improve the implementation in this PR:

### 1. Rethrow `CancellationException` in `RefreshableViewModel`
The `runCatching { task() }` block swallows all exceptions, including `CancellationException`. Swallowing `CancellationException` is an anti-pattern in Kotlin coroutines because it interferes with structured cancellation. If a ViewModel is cleared or a task is cancelled, the parent coroutine needs to know.

**Suggested change in `RefreshableViewModel.kt`**:
```kotlin
tasks.forEach { task ->
    launch {
        try {
            task()
        } catch (e: CancellationException) {
            throw e // Let coroutine cancel properly
        } catch (e: Exception) {
            Log.w(TAG, "refresh failed", e)
        }
    }
}
```

### 2. Manage/Cancel Concurrent Refresh Jobs
If a user quickly pulls to refresh multiple times, or switches selected years while a refresh is in progress, the app will execute redundant network requests concurrently. We should track the refresh job and cancel any active previous refresh before starting a new one.

**Suggested change in `RefreshableViewModel.kt`**:
```kotlin
abstract class RefreshableViewModel : ViewModel() {
    private val _isRefreshing = MutableStateFlow(false)
    val isRefreshing: StateFlow<Boolean> = _isRefreshing.asStateFlow()

    private var refreshJob: Job? = null

    protected fun refreshing(vararg tasks: suspend () -> Unit) {
        refreshJob?.cancel() // Cancel previous job to avoid waste/race conditions
        refreshJob = viewModelScope.launch {
            _isRefreshing.value = true
            try {
                coroutineScope {
                    tasks.forEach { task ->
                        launch {
                            try {
                                task()
                            } catch (e: CancellationException) {
                                throw e
                            } catch (e: Exception) {
                                Log.w(TAG, "refresh failed", e)
                            }
                        }
                    }
                }
            } finally {
                _isRefreshing.value = false
            }
        }
    }
}
```

### 3. Encapsulate the `isRefreshing` Guard inside `StateContent`
Every screen calling `StateContent` currently specifies `isRefreshing = isRefreshing && uiState !is UiState.Loading`. Since we never want to show the pull-to-refresh spinner when we are in the initial full-screen loading state, we can move this logic into `StateContent` so callers can just pass `isRefreshing = isRefreshing`.

**Suggested change in `StateContent.kt`**:
```kotlin
@Composable
fun <T> StateContent(
    state: UiState<T>,
    isRefreshing: Boolean,
    onRefresh: () -> Unit,
    // ...
) {
    PullToRefreshBox(
        isRefreshing = isRefreshing && state !is UiState.Loading, // Encapsulated check
        onRefresh = onRefresh,
        modifier = modifier,
    ) {
        // ...
    }
}
```
This simplifies call sites from:
```kotlin
isRefreshing = isRefreshing && uiState !is UiState.Loading
```
to:
```kotlin
isRefreshing = isRefreshing
```

gregmarra:
Thanks — applied 1 and 3, holding off on 2. Pushed in 3e7fbf197.

**1. Rethrow `CancellationException`** — done. Replaced `runCatching` with an explicit `try { task() } catch (e: CancellationException) { throw e } catch (e: Exception) { Log.w(...) }` so structured cancellation propagates.

**3. Encapsulate the `isRefreshing` guard** — done. `StateContent` now computes `isRefreshing && state !is UiState.Loading` internally, and all four call sites pass the raw `isRefreshing` flag.

**2. Cancel previous refresh job** — declining for this PR. A single VM-wide `refreshJob` is too coarse: several ViewModels have independent refresh entry points (e.g. `TeamDetailViewModel` has `refreshAll()`, `refreshTab(n)`, and `refreshYearData()`), so one shared job means an unrelated `refreshing()` call would cancel an in-flight one — a behavior change across ~8 ViewModels with real regression surface. The current concurrent refreshes are wasteful-but-idempotent (they write the same data), so there's no correctness bug to fix here. This is a behavioral enhancement beyond the scaffold's scope; happy to do it as a focused follow-up, scoped per-entry-point (or a keyed job map) rather than one shared job.

Separately, this push also fixes the `android-lint` CI failure: `ErrorBox`'s `modifier` is now the first optional parameter (Compose `ModifierParameter` lint rule).

gregmarra:
## Visual no-op verification

Built `main` (`94a158625`) and this branch, installed each to the same emulator (local backend, full 2026 data), and captured before/after for the three migrated list screens.

**Reproduction** (deterministic: signed out, fresh launch, top-of-list; Events auto-scrolls to Offseason in both builds)
- **Teams**, **Events**, **Districts** bottom-nav tabs

**Pixel diff** (status bar cropped):

| Screen | Differing pixels |
| --- | --- |
| Teams | **0 / 2,505,600** |
| Events | **0 / 2,505,600** |
| Districts | **0 / 2,505,600** |

Confirms the `UiState<T>` + `StateContent` migration leaves the **loaded (success) state** rendering byte-for-byte unchanged. (Scope note: this captures the success state; the shared loading / empty / error paths weren't pixel-captured.)

_Verified locally via emulator screenshot diff._

PR #1364, opened 2026-05-30
RFC: Extract shared :core module so :wear stops forking :app

gregmarra:
## RFC: Extract a shared `:core` module so `:wear` stops forking `:app`

**Status:** First migration step (`:core-network`, DTO layer) implemented in this PR. The remaining steps are follow-ups, one PR each — plan at the bottom.

### Motivation
`:wear` re-implements the networking + DTO stack that already exists in `:app`, with no shared module:
- `WearTbaApi` / wear DTOs duplicate `TbaApi` / `data.remote.dto`.
- `WearNetworkModule` duplicates `NetworkModule`.
- The copies have **already silently diverged** — e.g. `MediaDto.base64Image`: `:app` reads a top-level field, `:wear` reads it nested under `details`. Disk cache is **5 MB (wear) vs 20 MB (app)**; `setRequiresBatteryNotLow(true)` is on wear's periodic work but not app's.

### What this PR lands
The leaf-most shared layer — the DTOs — extracted into a real module both apps depend on:

- **New `:core-network` Android library** (Kotlin + serialization only; no Android-framework, no Hilt → low blast radius).
- **Moved `data.remote.dto` (18 files) into it**, package name preserved. Because the whole package moves together, `:app` needs **zero import changes** and there are **no split packages**.
- **Reconciled `MediaDto`** — the divergence above. The shared version exposes one flat accessor that prefers the top-level field then falls back to `details["base64Image"]`, which reproduces both modules' prior behavior exactly:
  ```kotlin
  val base64Image: String?
      get() = rawBase64Image ?: details?.get("base64Image")?.jsonPrimitive?.content
  ```
- **Deleted the 4 forked wear DTOs**; repointed `WearTbaApi` and the complication worker at the shared package.

Net: **+60 / −206**. `:wear` now consumes the canonical DTOs.

### Testing
- `./gradlew :core-network:assembleDebug :app:assembleDebug :wear:assembleDebug` — all green (Hilt aggregates across the new module).
- `./gradlew :app:testDebugUnitTest` — green.
- `./gradlew ktlintCheck` — clean across `:core-network`, `:app`, `:wear`.
- Phone-emulator smoke test: app launches (DI graph intact), Events list, team search + detail, and the Media tab all deserialize through the relocated DTOs without crashing.

### Migration plan (follow-up PRs)
1. **`:core-network`, API + provisioning** — move `TbaApi` into core and provide one OkHttp/Retrofit/Json from a shared Hilt module (base URL + API key injected per app via a small interface). Deletes `WearTbaApi` / `WearNetworkModule` and **fixes the cache / battery divergence** in one place. (Deferred here because it needs Hilt in the module + graph changes in both apps — kept separate so this DTO move stays a trivial, green diff.)
2. **`:core-domain`** — shared tracking algorithms (`findCurrentEvent`, record/RP computation, match selection) that `TeamTrackingComplicationWorker` and the app widget worker currently duplicate.
3. **`:core-data`** — one tracked-team store, removing the Glance-DataStore vs `TeamTrackerPreferences` dual-write.


gregmarra:
Here are a few suggestions to improve the implementation in this PR:

### 1. Clean up Redundant Mapping Logic
In `Mappers.kt` (under `app/src/main/kotlin/com/thebluealliance/android/data/mappers/Mappers.kt`), we currently perform the fallback check for `base64Image` when mapping to `MediaEntity`:
```kotlin
base64Image =
    base64Image
        ?: details?.get("base64Image")?.let {
            (it as? kotlinx.serialization.json.JsonPrimitive)?.content
        },
```
Since the newly-shared `MediaDto` now encapsulates this fallback in its getter, this block can be simplified directly to:
```kotlin
base64Image = base64Image
```

### 2. Enhance Null/Type Safety of `MediaDto.base64Image` Getter
In `MediaDto.kt` (under `core-network/src/main/kotlin/com/thebluealliance/android/data/remote/dto/MediaDto.kt`), the custom getter is implemented as:
```kotlin
val base64Image: String?
    get() = rawBase64Image ?: details?.get("base64Image")?.jsonPrimitive?.content
```
There are two minor edge-case risks here:
* If the API returns a JSON null (`"base64Image": null`) inside the `details` object, `details?.get("base64Image")` returns a `JsonNull` object. Accessing `.content` on it will return the string `"null"`, which will cause issues when the app tries to render it.
* If the API returns a malformed structure (e.g. an array or nested object under `"base64Image"`), `.jsonPrimitive` will throw an `IllegalArgumentException`.

We can make this safer and more robust by using `as? JsonPrimitive` and `contentOrNull`:
```kotlin
val base64Image: String?
    get() = rawBase64Image ?: (details?.get("base64Image") as? JsonPrimitive)?.contentOrNull
```

### 3. Align Kotlin Compiler JVM Target
In `core-network/build.gradle.kts`, Java 17 compatibility is set inside the `android` block via `compileOptions`, but the Kotlin compiler JVM target is not explicitly set. To prevent potential mismatched bytecode warnings or compiler warnings during build, configure `kotlinOptions`:
```kotlin
android {
    // ...
    compileOptions {
        sourceCompatibility = JavaVersion.VERSION_17
        targetCompatibility = JavaVersion.VERSION_17
    }

    kotlinOptions {
        jvmTarget = "17"
    }
}
```

gregmarra:
Thanks — applied 1 and 2, holding off on 3. Pushed in 881145be6.

**1. Redundant mapping logic** — done. `MediaDto.toEntity` now just does `base64Image = base64Image`, since the DTO getter encapsulates the top-level/nested fallback.

**2. Null/type safety of the getter** — done. Switched to `rawBase64Image ?: (details?.get("base64Image") as? JsonPrimitive)?.contentOrNull`, so a JSON `null` or a malformed array/object under `details["base64Image"]` now yields `null` instead of the string `"null"` or an `IllegalArgumentException`.

**3. Align Kotlin `jvmTarget`** — declining. This project is on AGP 9.2.1 with built-in Kotlin, and AGP already aligns the Kotlin `jvmTarget` to `compileOptions.targetCompatibility` (17) automatically — I rebuilt `:core-network` with `--rerun-tasks` and there's no JVM-target/bytecode-mismatch warning. No other module (`:app`, `:wear`, …) sets `jvmTarget` either, so adding a `kotlinOptions { jvmTarget = "17" }` block to only `:core-network` would be redundant and inconsistent (and `kotlinOptions {}` is the legacy DSL under AGP 9's built-in Kotlin). Happy to set it project-wide via a convention if we ever want it explicit, but that'd be its own change.

PR #1362, opened 2026-05-30
RFC: Quick-wins sweep — remove dead weight & dedup

gregmarra:
## RFC: Quick-wins sweep — dead weight & dedup

**Status:** Implemented in this PR (a focused, low-risk slice). Remaining candidates listed at the bottom.

### Motivation
The app has accumulated small duplications and a little dead code that make routine edits noisier than they need to be. None of these are bugs — each is a paper cut. This RFC bundles the safest, highest-frequency ones into a single mechanical, **no-behavior-change** sweep, and establishes the named helpers so the patterns don't re-spread.

### What this PR does

**1. Delete dead code**
`EventRepository.observeTeamEventKeys(teamKey)` had zero callers. Removed.

**2. `String.teamNumber` extension** — `util/TeamKey.kt`
The transform "FRC team key → team number" (`"frc254" → "254"`) was open-coded ~30 times as either `key.removePrefix("frc")` or, inconsistently, `key.substring(3)`. Replaced every site with one named extension:
```kotlin
val String.teamNumber: String get() = removePrefix("frc")
```
Swept 12 files (event tabs, match list/detail, district/regional screens, the home-screen widget + worker, pit-map URLs). Folding the positional `substring(3)` sites in also makes them robust to keys that don't begin with `frc`.

**3. `Ranking.recordString` extension** — `domain/model/Ranking.kt`
`"${r.wins}-${r.losses}-${r.ties}"` was inlined in 4 places. Replaced with:
```kotlin
val Ranking.recordString: String get() = "$wins-$losses-$ties"
```

**4. Consolidate `RpDots`**
The ranking-point bonus-dots composable was duplicated (near-identical ~28-line copies) in `MatchDetailScreen` and `MatchList`, differing only in dot size / spacing / padding. Extracted one parameterized `ui/components/RpDots.kt`; the two call sites pass their own sizes via params + modifier, so the rendering is byte-for-byte unchanged.

### Result
15 existing files touched + 2 tiny new helper files; net **−42 lines** of app code. Pure refactor, no behavior change.

### Testing
- `./gradlew :app:testDebugUnitTest` — green (existing suite, incl. `EventRepositoryTest`)
- `./gradlew :app:ktlintCheck` — clean

### Deferred (intentionally out of scope here)
Originally floated for this sweep, but larger or riskier — each better as its own change so this diff stays trivially reviewable:
- Remove stale legacy module directories
- A Gradle build-convention plugin (not a "quick win"; warrants its own RFC)
[...]

PR #1366, opened 2026-05-30
refactor: add String.teamNumber extension and replace open-coded sites

gregmarra:

The transform "FRC team key → team number" (`"frc254" → "254"`) was open-coded ~30 times as either `removePrefix("frc")` or, inconsistently, `substring(3)`. Replace every site with one named extension in `util/TeamKey.kt`:

```kotlin
val String.teamNumber: String get() = removePrefix("frc")
```

Swept across 12 files: event tabs (rankings/insights/alliances/awards/advancement), match list + detail, district + regional screens, the home-screen widget + worker, and pit-map URLs. Folding the positional `substring(3)` sites in also makes them robust to keys that don't begin with `frc`.

No behavior change.

### Testing

### DIFF
--- a/app/src/main/kotlin/com/thebluealliance/android/ui/components/MatchList.kt
+++ b/app/src/main/kotlin/com/thebluealliance/android/ui/components/MatchList.kt
@@ -39,6 +39,7 @@ import com.thebluealliance.android.domain.model.Match
 import com.thebluealliance.android.domain.model.PlayoffType
 import com.thebluealliance.android.domain.model.displayTitleShort
 import com.thebluealliance.android.domain.rpBonuses
+import com.thebluealliance.android.util.teamNumber
 import kotlinx.coroutines.launch
 
 @OptIn(ExperimentalFoundationApi::class)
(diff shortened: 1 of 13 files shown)


gregmarra:
This is a nice readability improvement. Using `teamKey.teamNumber` makes the code significantly more self-documenting than raw string manipulation implementations like `removePrefix("frc")` and `substring(3)`.

One recommendation to get even more value from this:

### Move `TeamKey.kt` to the shared `:core-network` module
Currently, `TeamKey.kt` is added to the `app` module. However, the Wear OS module still performs manual string manipulation to parse team keys in `TeamTrackingComplicationWorker.kt`:
```kotlin
// In wear's TeamTrackingComplicationWorker.kt
private fun teamKeysToNumbers(keys: List<String>?): String =
    keys?.joinToString(", ") { it.removePrefix("frc") } ?: ""
```

Since the shared `:core-network` module (introduced in #1364) is imported by both `:app` and `:wear`, moving `TeamKey.kt` to `:core-network` (e.g. under package `com.thebluealliance.android.core.network.util`) will allow both modules to reuse the extension property and eliminate the remaining manual `removePrefix("frc")` calls.

gregmarra:
Agreed on the end state — Wear's `TeamTrackingComplicationWorker` still hand-parses team keys with `removePrefix("frc")`, and that duplication should go away.

The blocker is sequencing: `:core-network` doesn't exist on `main` yet — it's introduced by #1364, which isn't merged. This branch is cut from `main`, so moving `TeamKey.kt` into `:core-network` here would couple this PR to #1364 and break its independence (the whole point of keeping these as separate, individually-mergeable cleanups).

So I'll keep `TeamKey.kt` in `:app` for this PR, and do the move to `:core-network` + the Wear dedup as a focused follow-up once #1364 lands. Declining the move here.

PR #1367, opened 2026-05-30
refactor: add Ranking.recordString extension

gregmarra:

`"${r.wins}-${r.losses}-${r.ties}"` was inlined in several places. Introduce one extension in `domain/model/Ranking.kt`:

```kotlin
val Ranking.recordString: String get() = "$wins-$losses-$ties"
```

and use it in the event rankings table and the team-event summary row.

> **Behavior note:** in the rankings table, the primary/secondary sort columns now fall back to the W-L-T record when a sort-order value is missing, where they previously showed `--`. This is a small visible change, not a pure dedup — called out here so it's reviewed deliberately. Happy to drop it if you'd rather keep `--`.

### Testing

### DIFF
--- a/app/src/main/kotlin/com/thebluealliance/android/domain/model/Ranking.kt
+++ b/app/src/main/kotlin/com/thebluealliance/android/domain/model/Ranking.kt
@@ -13,3 +13,7 @@ data class Ranking(
     val sortOrders: List<Double> = emptyList(),
     val extraStats: List<Double> = emptyList(),
 )
+
+/** Win-loss-tie record formatted as "W-L-T", e.g. "10-2-0". */
+val Ranking.recordString: String
+    get() = "$wins-$losses-$ties"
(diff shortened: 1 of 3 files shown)


gregmarra:
Extracting `Ranking.recordString` as an extension property is a nice minor cleanup. However, there's a search-and-replace bug in `EventRankingsTab.kt` that was introduced in this PR:

### Revert fallback change for primary/secondary sort values
In `EventRankingsTab.kt`, the fallback when a sort order is missing or null was changed from `?: "--"` to `?: ranking.recordString`:

```kotlin
val primarySortValue =
    ranking.sortOrders.getOrNull(0)?.let { value ->
        val precision = sortOrders?.getOrNull(0)?.precision ?: 2
        String.format(Locale.US, "%.${precision}f", value)
    } ?: ranking.recordString // Here
```

This is incorrect because:
1. If an event has no sort orders (or they haven't loaded yet), the table will display the team's Win-Loss-Tie record (e.g. `10-2-0`) inside the "Ranking Score" and "Coopertition" columns.
2. The team's record is already displayed in the "Record" column immediately next to them, leading to confusing duplicate values under wrong column headers.

Please revert those two fallback changes back to the original `?: "--"`. Only the actual record text view should be updated to use `ranking.recordString`.

gregmarra:
Good catch — fixed in fd1e10d17. Reverted both the primary and secondary sort-value fallbacks back to `?: "--"`. The only remaining behavior change vs `main` is the Record column itself now using `ranking.recordString` (which is identical output to the old `"${wins}-${losses}-${ties}"`, just via the new extension). So the duplicate-record-under-wrong-columns issue is gone.

gregmarra:
I don't think this is worth the indirection given how simple the formatting is and it's only in two places right now.

PR #1368, opened 2026-05-30
refactor: consolidate duplicate RpDots composable

gregmarra:

The ranking-point bonus-dots composable was duplicated as near-identical ~28-line copies in `MatchDetailScreen` and `MatchList`, differing only in dot size / spacing / padding. Extract one parameterized `ui/components/RpDots.kt`:

```kotlin
@Composable
fun RpDots(
    bonuses: List<Boolean>,
    achievedColor: Color,
    modifier: Modifier = Modifier,
    dotSize: Dp = 6.dp,
    gap: Dp = 2.dp,
)
```

The two call sites pass their own sizes via params + modifier, so the rendering is byte-for-byte unchanged. No behavior change.

### Testing

### DIFF
--- a/app/src/main/kotlin/com/thebluealliance/android/ui/components/MatchList.kt
+++ b/app/src/main/kotlin/com/thebluealliance/android/ui/components/MatchList.kt
@@ -1,6 +1,5 @@
 package com.thebluealliance.android.ui.components
 
-import androidx.compose.foundation.Canvas
 import androidx.compose.foundation.ExperimentalFoundationApi
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Arrangement
@@ -29,8 +28,6 @@ import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.graphics.Color
-import androidx.compose.ui.graphics.drawscope.Stroke
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.unit.dp
 import com.thebluealliance.android.domain.getGroup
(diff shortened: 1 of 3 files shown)


gregmarra:
This is a great cleanup. Consolidating the duplicate implementations of `RpDots` into a single parameterized component under components makes it significantly easier to maintain and reuse.

One recommendation to improve the component's flexibility:

### Expose a parameter for the unachieved dot color
Inside `RpDots.kt`, the border color for unachieved/hollow ranking points is currently hardcoded as `Color(0xFF9CA3AF)`:

```kotlin
drawCircle(
    color = Color(0xFF9CA3AF),
    radius = size.minDimension / 2 - 1.dp.toPx(),
    style = Stroke(width = 1.dp.toPx()),
)
```

To make the component fully flexible (e.g., if different card background colors or future dark mode adjustments require a different outline contrast), we should expose this as a parameter. You can default it to the original color so that existing call sites don't need any changes:

```kotlin
@Composable
fun RpDots(
    bonuses: List<Boolean>,
    achievedColor: Color,
    modifier: Modifier = Modifier,
    dotSize: Dp = 6.dp,
    gap: Dp = 2.dp,
    unachievedColor: Color = Color(0xFF9CA3AF), // Exposed parameter
) {
    // ...
    drawCircle(
        color = unachievedColor,
        radius = size.minDimension / 2 - 1.dp.toPx(),
        style = Stroke(width = 1.dp.toPx()),
    )
}
```
Alternatively, defaulting it to `MaterialTheme.colorScheme.outline` would align it nicely with the app's M3 theme tokens.

gregmarra:
Going to hold off on this one. Both original `RpDots` implementations used the same hardcoded `Color(0xFF9CA3AF)`, so this PR's job is purely to dedupe identical code — preserving that exact color keeps it a no-visual-change refactor, which is what I want for a consolidation PR.

Exposing `unachievedColor` now would add a parameter no call site uses (YAGNI), and switching the default to `MaterialTheme.colorScheme.outline` would silently change the rendered outline color — turning a pure refactor into a behavior change. I'd rather expose it the moment a real need shows up (e.g. dark-mode contrast on a different card background), at which point the right default becomes obvious. Declining for now.

gregmarra:
## Visual no-op verification

Built `main` (`94a158625`) and this branch, installed each to the same emulator (pointed at a local backend with full 2026 data), and captured before/after screenshots at the two sites this PR touches — both render `RpDots`.

**Reproduction** (deterministic: same backend data + identical navigation)
- Match list — deep-link `https://www.thebluealliance.com/event/2026cthar` → **Matches** tab
- Match detail — deep-link `https://www.thebluealliance.com/match/2026cthar_qm1`

**Pixel diff** (status bar cropped so the clock tick doesn't count):

| Screen | Differing pixels |
| --- | --- |
| Match list | **0 / 2,505,600** |
| Match detail | **0 / 2,505,600** |

The single consolidated `RpDots` renders pixel-for-pixel identically to the two duplicates it replaces.

_Verified locally via emulator screenshot diff._

PR #1369, opened 2026-05-30
RFC: Typed per-screen Tab enums for refreshTab(...)

gregmarra:
## RFC: Typed per-screen `Tab` enums for `refreshTab(...)`

**Status:** Draft / Request for Comments — RFC doc only, no implementation in this PR.

### Problem

Four detail ViewModels still take a raw `Int` page index and dispatch with `when (tab) { 0 -> …, 1 -> … }` plus `// Info` / `// Events` comments. The screen-side caller passes the `HorizontalPager` page number directly, and each screen independently declares a parallel `private val TABS = listOf("…")`. The two have to stay in lockstep but nothing enforces it.

Raw-`Int` sites:
- `DistrictDetailViewModel.kt:117` ↔ `DistrictDetailScreen.kt:56` (2 tabs)
- `TeamDetailViewModel.kt:203` ↔ `TeamDetailScreen.kt:78` (3 tabs)
- `MyTBAViewModel.kt:139` ↔ `MyTBAScreen.kt:74` (2 tabs)
- `TeamEventDetailViewModel.kt:256` ↔ `TeamEventDetailScreen.kt:70` (5 tabs)

Found while auditing: `TeamEventDetailViewModel.refreshTab` only handles `0..3` but `TABS` has 5 entries — **pulling to refresh on the Awards tab silently does nothing today.**

### Precedent already in the repo

`EventDetailViewModel.refreshTab(tab: EventDetailTab)` (file:`EventDetailViewModel.kt:377`) already takes a typed enum (`EventDetailScreen.kt:72` defines `enum class EventDetailTab`), and the screen calls `refreshTab(EventDetailTab.entries[page])`. We just need to do the same thing in the other four screens.

### Proposal

For each remaining screen: add an `enum class <Screen>Tab` next to the existing `TABS` list, retype `refreshTab(tab: <Screen>Tab)`, and have the screen pass `<Screen>Tab.entries[page]`. Exhaustive `when` over the enum then prevents the next missing-case bug. Fold in the missing `AWARDS` refresh while we're in `TeamEventDetail`.

Full doc: [`rfcs/0005-typed-tab-enums.md`](../blob/rfc/typed-tab-enums/rfcs/0005-typed-tab-enums.md)

### Decision requested

Approve the direction (typed per-screen enums matching `EventDetailTab`). On approval, implementation lands as a follow-up PR (or four — they don't depend on each other).

---

Spotted while reviewing #1363's `refreshTab(Int)` sites; split out per the RFC-per-item workflow. The implementation PR(s) will conflict with #1363 since they touch the same files — we'll rebase that out separately.

### DIFF
--- a/rfcs/0005-typed-tab-enums.md
+++ b/rfcs/0005-typed-tab-enums.md
@@ -1,0 +1,156 @@
+# RFC 0005: Typed per-screen `Tab` enums for `refreshTab(...)`
+
+- **Status:** Draft / Request for Comments
+- **Author:** @gregmarra
+- **Date:** 2026-05-30
+- **Type:** Refactor (mechanical, contained)
+
+## Summary
+
+Four detail ViewModels still take a raw `Int` page index in `refreshTab(tab: Int)`
+and dispatch with `when (tab) { 0 -> …, 1 -> … }`. The screen-side caller passes
+the `HorizontalPager` page number directly. The mapping of "0 means …, 1 means
+…" lives only in `//` comments next to each case. Replace each with a typed
+per-screen enum so the dispatch becomes exhaustive and the meaning of each
+position is named.
+
+The precedent already exists in this repo: `EventDetailViewModel.refreshTab(tab:
+EventDetailTab)` does exactly this — its screen calls
+`viewModel.refreshTab(EventDetailTab.entries[page])`.
+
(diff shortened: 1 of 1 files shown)


gregmarra:
Spec-only RFC — moved to #1370. The markdown isn't something we'd want to merge; the implementation will land in a future PR that references the issue.

PR #1387, opened 2026-06-10
Opt in to Certificate Transparency across app/tv/wear (Android 17 readiness)

gregmarra:
## Summary
- Android 17 (API 37) enables Certificate Transparency by default for apps targeting it; the `<certificateTransparency>` element is opt-in on API 36 and ignored below. Opt in now in all three modules' network security configs so we get CT enforcement ahead of the (deliberate, post-stable) targetSdk bump.
- Unifies the debug-cleartext pattern on resource shadowing, which `:wear` already used: `:app` and `:tv` now declare the config in their **main** manifests and let the debug source set shadow the `network_security_config.xml` resource. That removes `:app`'s `tools:replace` overlay and deletes `:tv`'s debug-only manifest entirely.

## Panel notes
- **Android Framework Nerd / Tech Lead:** straight from the Android 17 migration guidance; release builds previously had *no* network security config in `:app`/`:tv`, so this also makes "cleartext off" explicit rather than implicit.

## Test plan
- [x] `./gradlew :app:assembleDebug :tv:assembleDebug :wear:assembleDebug` green
- [x] `:app:testDebugUnitTest` green
- [x] Debug app installs and runs on an API 37 (Android 17 Beta) emulator
- [ ] CI lint (warningsAsErrors) on all modules


gregmarra:
Verification: all three modules assemble green; debug app installs and launches on the API 37 (Android 17 Beta 4) emulator running the new `android-37.0` Play-Store 16k-page image — no NSC parse issues on boot. CT enforcement itself activates on API 36+ devices per the platform docs.

PR #1390, opened 2026-06-10
Display match times in the event's timezone

gregmarra:
## Summary
- Match rows (`MatchList`) and the match detail header formatted times with `ZoneId.systemDefault()`, so anyone following a remote event saw times that disagreed with the venue schedule, TBA web, and FRC schedules — a parent in California saw "Sat 6:00a" for a 9:00 AM East Coast match.
- `Event.timezone` was already stored on the entity/domain model but unused by any UI. It now drives both formatters (match list rows, team@event summary rows, match detail header), falling back to device time when missing or invalid.
- The home-screen widget intentionally keeps device-local time for its at-event use case.

## Panel notes
- **PM:** "small fix with outsized trust impact during competition season" — their top quick win.
- **FRC Team Member:** FRC schedules speak venue-local; the app should too.

## Test plan
- [x] New `MatchTimeFormatTest` covers explicit zones (NY/Tokyo), invalid-zone fallback, and null time
- [x] `:app:assembleDebug` + `:app:testDebugUnitTest` green
- [ ] Before/after screenshots of a match list (PR comment to follow)

### DIFF
--- a/app/src/main/kotlin/com/thebluealliance/android/ui/components/MatchList.kt
+++ b/app/src/main/kotlin/com/thebluealliance/android/ui/components/MatchList.kt
@@ -45,6 +45,7 @@ fun MatchList(
     matches: List<Match>?,
     playoffType: PlayoffType,
     onNavigateToMatch: (String) -> Unit,
+    eventTimezone: String? = null,
     headerContent: (LazyListScope.() -> Unit)? = null,
     headerItemCount: Int = 0,
     innerPadding: PaddingValues = PaddingValues.Zero,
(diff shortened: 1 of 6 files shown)


gregmarra:
Verification: `MatchTimeFormatTest` pins explicit zones (`America/New_York` → "Wed 7:00p", `Asia/Tokyo` → "Thu 9:00a" for the same instant), invalid-zone fallback to device time, and the null em-dash; `:app:testDebugUnitTest` + `:app:assembleDebug` green. Emulator tz is America/Los_Angeles, so before/after on an ET event will show the 3-hour correction once the dev backend is back.

gregmarra:
I don't actually like this, I think having all the times displayed in the phone's time zone makes more sense if you are not physically present, and if you are physically present it is moot.

PR #1400, opened 2026-06-10
Page the teams refresh until empty instead of hardcoded pages 0-19

gregmarra:
## Summary
- `TeamsViewModel.refreshTeams()` fired exactly pages 0–19; the API pages by 500 team numbers, so page 19 ends at frc9999 — teams numbered 10000+ went permanently stale after first install, and brand-new rookies never appeared in the Teams tab.
- New `TeamRepository.refreshAllTeams()` loops `refreshTeamsPage` until an empty page (40-page safety cap ≈ frc19999 headroom), and both `TeamsViewModel` and `DataSyncManager.syncTeams()` now share it — previously two divergent paging strategies, one silently wrong.
- Note: the per-page parallel fan-out is gone; loop-until-empty is inherently sequential (you can't know where the list ends in advance), and the user-facing refresh still completes in a few seconds over ~30 small cached requests.

## Panel notes
- **Tech Lead:** "we even have the right primitive — refreshTeamsPage returns the count — and the ViewModel ignores it. Classic copy-paste divergence."
- **FRC Team Member:** rookie numbers passed 10000 seasons ago; this matters every kickoff.

## Test plan
- [x] New `TeamRepositoryTest`: pages until the empty page (verifies stop + no over-fetch) and respects the safety cap
- [x] `:app:testDebugUnitTest` green


app/src/main/kotlin/com/thebluealliance/android/data/repository/TeamRepository.kt
```diff
@@ -101,4 +112,9 @@ class TeamRepository
             } catch (_: Exception) {
             }
         }
+
+        private companion object {
+            // 40 pages x 500 team numbers covers keys up to frc19999.
+            const val MAX_TEAM_PAGES = 40
```
bherbst: Why do we need this at all now? Feels like if `refreshTeamsPage()` is returning > 1 for non-existent pages of teams we are likely running into all sorts of other oddities anyway

bherbst (approved):

app/src/main/kotlin/com/thebluealliance/android/data/repository/TeamRepository.kt
gregmarra: Hey, this is Claude. Fair point, and let me separate the two things going on. The `> 0` empty-page check is the real terminator and handles the normal case — the cap isn't there because I expect `refreshTeamsPage()` to lie about non-existent pages. `MAX_TEAM_PAGES` is purely a bounded-loop guard so a misbehaving *remote* (a page that never reports empty) can't turn a background sync into an unbounded fetch loop hammering the API.

That said, you've caught the real smell: 40 is a magic number that would *silently* truncate at frc19999, which is the wrong failure mode. Two honest options — drop the cap and trust the empty-page terminator, or keep the guard but `Log.w` when we actually hit it so truncation is loud instead of silent. I lean toward the loud guard, but it's your call.

PR #1405, opened 2026-06-11
Search: honest empty states instead of a blank screen

gregmarra:
## Summary
- Search rendered pure background both before typing and when a query matched nothing — reachable from nearly every screen's top bar, and users couldn't tell loading from broken from empty.
- Blank query → "Search teams and events" hint. Settled no-match query → `No results for "<query>"`.
- The no-results check compares the live query against the **debounced** query (`searchedQuery`, newly carried in `SearchUiState`), so the message can't flash mid-keystroke while results lag the 300ms debounce.

## Test plan
- [x] `:app:compileDebugKotlin` green (only pre-existing `AndroidGradlePluginVersion` lint warning-as-error, unrelated to this change)
- [ ] Live emulator verification of both states (not re-run since the message simplification)

## Screenshots

| | Before | After |
|---|---|---|
| **Search with no results** |  |  |
| **Long query stays within the margins (padding fix)** |  |  |


bherbst (approved):
That screenshot looks like `EmptyBox` could use a bit of padding, but looks good with respect to this specific set of changes

gregmarra:
Hey, this is Claude. Good eye — agreed, `EmptyBox` sits a little tight against the edges. Since it's a shared component I kept it out of this PR's scope (this one's just about replacing the blank screen with honest empty states), but the padding is worth a quick follow-up and I'll open one.

PR #1408, opened 2026-06-11
fix(ui): sync system bar styling with the in-app theme override

gregmarra:
## Summary
- `enableEdgeToEdge` was called once in `MainActivity.onCreate` with `SystemBarStyle.auto` for the navigation bar, which resolves against the *system* night mode — it never sees the in-app `ThemeMode` override from `ThemePreferences`.
- With in-app DARK on a system-light device using 3-button nav, the nav bar got a light auto-scrim with dark icons over the near-black `NavigationBar`, leaving a mismatched band (and vice versa for in-app LIGHT on a system-dark device).
- Fix follows the standard Now-in-Android recipe: `TBAApp` (which resolves the effective `darkTheme`) re-invokes `enableEdgeToEdge` in a `DisposableEffect(darkTheme)` on the host activity via `LocalActivity`, passing `SystemBarStyle.auto(TRANSPARENT, TRANSPARENT) { darkTheme }` for the nav bar. The status bar stays `SystemBarStyle.dark` since the top app bar is always TBABlue. `MainActivity`'s initial call remains as the pre-compose default.

## Panel notes
Android Framework Nerd flagged that the one-shot `enableEdgeToEdge` snapshot in `onCreate` can never track the Compose-resolved theme, so `SystemBarStyle.auto`'s detector disagrees with the rendered UI whenever the in-app override diverges from system night mode.

## Test plan
- [x] `./gradlew :app:lintDebug` passes
- [ ] Visual verification on emulator (3-button nav, in-app DARK on system-light) — handled by orchestrator


gregmarra:
Visually verified on the API 37 emulator (3-button nav enabled, in-app theme set to Dark while the system is light): before this PR the nav bar rendered as a light-gray band with dark icons under the dark UI; with this PR it matches the dark theme (dark bar, light icons). Before/after screenshots: `1408_navbar_before.png` / `1408_navbar_after.png` in the improvement report.

app/src/main/kotlin/com/thebluealliance/android/ui/TBAApp.kt
```diff
@@ -50,6 +56,18 @@ fun TBAApp(
...
+        // Keep system bars in sync with the in-app theme override, not just the system
+        // night mode; the status bar stays dark-style because the top bar is always TBABlue.
+        activity?.enableEdgeToEdge(
+            statusBarStyle = SystemBarStyle.dark(Color.TRANSPARENT),
+            navigationBarStyle =
+                SystemBarStyle.auto(Color.TRANSPARENT, Color.TRANSPARENT) { darkTheme },
+        )
+        onDispose {}
+    }
+
```
bherbst, line 69: Why a `DisposableEffect` instead of a `LaunchedEffect` if we don't care about `onDispose`?

bherbst (approved):

app/src/main/kotlin/com/thebluealliance/android/ui/TBAApp.kt
gregmarra: Hey, this is Claude. You're right — there's nothing to dispose here, so `DisposableEffect` is overkill. `enableEdgeToEdge` is a synchronous one-shot call with no teardown, so `LaunchedEffect(darkTheme)` actually expresses the intent better: re-run when the theme flips, nothing to undo. I reached for `DisposableEffect` out of habit, since imperative window/system-bar APIs *usually* want cleanup — but this one genuinely doesn't. I'll switch it to `LaunchedEffect`.

Issue #1391, opened 2026-06-10
myTBA: show team/event names instead of raw keys, link Match rows, actionable empty states


gregmarra:
## Problem

The myTBA screen renders raw API keys instead of human-readable names, and Match rows are dead taps:

- A favorited team renders as `frc254 / Team` and an event as `2026casj / Event`, even though the display names ("254 — The Cheesy Poofs", "Silicon Valley Regional") are usually already in our Room tables (`MyTBAScreen.kt:393-396`, `503-506`).
- Subscription rows print raw server keys ("Team · upcoming_match, match_score") although `NotificationType.displayName` exists for exactly this (`MyTBAScreen.kt:508`).
- Match favorites/subscriptions synced from the web render tappable but the tap is a no-op — the `when` blocks only handle `ModelType.TEAM`/`EVENT` (`MyTBAScreen.kt:352-356`, `466-470`).
- Empty states say "No favorites yet" without telling the user how to add one (`MyTBAScreen.kt:343`, `457`).

## Proposed approach

- In `MyTBAViewModel`, combine the favorites/subscriptions flows with `TeamDao`/`EventDao` lookups to map keys → display names (fall back to the raw key when uncached; optionally fetch missing teams/events on refresh).
- Render notification types via `NotificationType.fromServerKey()?.displayName`.
- Add an `onNavigateToMatch` callback wired to the existing `Screen.MatchDetail` for `MATCH` rows.
- Empty-state copy: "Tap the star on any team or event to add it here."

## Why not now

Needs a small amount of data plumbing (key → name joins, uncached-name fetch policy) — more than a same-day change. Surfaced by the PM seat of the improvement-survey panel (2026-06-10); myTBA is the retention surface and currently reads like a debug dump.

Issue #826, opened 2017-03-12
Disabling myTBA is confusing


1fish2:
* Go to Settings.
* Turn off the "Enable myTBA" switch.
--> It immediately goes to the screen: "While you wait, why not check out myTBA?" It has actions CANCEL and SKIP INTRO.

While I wait for what? Cancel what?

gregmarra:
Closing as obsolete: the confusing "Enable myTBA" switch this issue describes no longer exists. myTBA is now managed via Google sign-in/sign-out on the myTBA screen, with an explicit "Sign out?" confirmation dialog; Settings contains only theme + notification rows. Remaining myTBA polish (raw keys instead of names, dead Match rows, empty-state guidance) is tracked in #1391.

PR #1411, opened 2026-06-11
fix(mytba): only apply local myTBA writes on confirmed server success

gregmarra:
## Summary
- myTBA writes only checked the body code for 401 — any other server failure still ran the local DAO insert/delete, desyncing the device from the backend: the star fills in, the user thinks they're subscribed, and the push never comes (relates to #826 / #1391).
- `MyTBARepository` now validates every ClientAPI response before touching Room: list endpoints must return a 2xx body code; `setPreferences` must return its documented success code **0** (verified verbatim against the backend's `client_api.py` — it nests per-model results in the `message` JSON, so the naive "require 2xx" fix would have rejected every *successful* favorite write). The 401 refresh-then-throw special case is preserved.
- Failed `refreshFavorites`/`refreshSubscriptions` responses (e.g. a 401/500 body with an empty list) no longer wipe the local cache inside the transaction.
- `toggleFavorite` in `TeamDetailViewModel`/`EventDetailViewModel` swallowed all of this silently while `updatePreferences` right next to it emitted a snackbar message; it now emits "Couldn't update favorite" through the same existing `_userMessage` flow — this also surfaces plain network failures (airplane mode, HTTP 5xx) that were previously invisible.

## Panel notes
Tech Lead: "the user thinks they're subscribed and never gets the push" — silent-success on non-401 errors plus a silent favorite star left no signal that a myTBA write was lost.

## Test plan
- [x] New `MyTBARepositoryTest` (11 cases): local write happens on confirmed success (including the real `code=0` setPreferences contract), throws and skips the DAO write on 500, 401 triggers the refresh-then-throw path, error responses never wipe the local cache, request carries the device key.
- [x] `./gradlew :app:testDebugUnitTest :app:lintDebug -q` — clean.

### DIFF
--- a/app/src/main/kotlin/com/thebluealliance/android/data/repository/MyTBARepository.kt
+++ b/app/src/main/kotlin/com/thebluealliance/android/data/repository/MyTBARepository.kt
@@ -13,8 +13,16 @@ import com.thebluealliance.android.domain.model.Favorite
 import com.thebluealliance.android.domain.model.Subscription
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.map
+import java.net.HttpURLConnection
 import javax.inject.Inject
 import javax.inject.Singleton
+
+private val HTTP_SUCCESS = 200..299
+
+// The setPreferences endpoint reports success as code 0 with per-model results
+// nested in the message JSON; the list endpoints use HTTP-style body codes
+// (see backend src/backend/api/handlers/client_api.py).
+private const val SET_PREFERENCES_SUCCESS = 0
 
 @Singleton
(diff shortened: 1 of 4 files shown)


gregmarra:
Orchestrator review note: the agent verified against the upstream backend that `setPreferences` returns `code=0` on success, so the success check deliberately accepts 0 **or** 2xx for updates (lists require 2xx). One inherited caveat for reviewers: `BaseResponseDto.code` defaults to 0 when the field is absent, so a malformed empty body would also read as update-success — pre-existing kotlinx-default semantics, unchanged by this PR; flagging for awareness.

PR #1418, opened 2026-06-11
fix(ui): grow inline link touch targets to the 48dp minimum

gregmarra:
## Summary
- Grows the app's undersized inline tap targets to the 48dp accessibility minimum using pure modifier work — no layout redesigns.
- **MatchDetailScreen**: the event link and the stacked alliance team numbers (previously ~24dp tall at a 2dp pitch) now use `clickable(...).minimumInteractiveComponentSize()`, giving each a >=48dp box and a 48dp pitch between team numbers.
- **EventRankingsTab**: the team-number link nested inside the expandable row now fills a full 48dp target instead of ~24dp, resolving the two competing targets; the row swaps its 12dp vertical padding for `heightIn(min = 48.dp)` so its visual height is unchanged.
- **EventInfoTab**: the address / website / pit-map / webcast link rows get `heightIn(min = 48.dp)` inside the clickable (rows were ~20dp tall). LazyColumn modifier and the week line are untouched (owned by other open PRs).
- **MoreScreen**: the bodySmall About/Thanks footer links (~16dp tall) get full 48x48 `minimumInteractiveComponentSize()` boxes, vertically centered against the dot separator.
- All touched links/rows gain `Role.Button` semantics.

## Panel notes
Designer — "the highest-traffic links in the app, used in gloves and bleachers, are the hardest things to hit."

## Test plan
- [x] `./gradlew :app:testDebugUnitTest :app:lintDebug` passes (warningsAsErrors)
- [x] Diff is modifier-only: clickable regions wrap the size expansion (`clickable` precedes `minimumInteractiveComponentSize`/`heightIn`), so the grown area is actually tappable
- [ ] orchestrator visual pass

## Screenshots

| | Before | After |
|---|---|---|
| **Event Info — inline link rows grow to 48dp touch targets** |  |  |

### DIFF
--- a/app/src/main/kotlin/com/thebluealliance/android/ui/events/detail/tabs/EventInfoTab.kt
+++ b/app/src/main/kotlin/com/thebluealliance/android/ui/events/detail/tabs/EventInfoTab.kt
@@ -7,6 +7,7 @@ import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.heightIn
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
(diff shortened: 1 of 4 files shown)


gregmarra:
Orchestrator verification (API 37 emulator, UI-tree measurements on the More screen): About/Thanks are now true 48×48dp targets (126×126px @420dpi) with real spacing between them — before, the reported expansion clamped/tiled (Thanks 126×95px, zero gap) — and both nodes' accessibility class changed TextView → Button via the Role.Button semantics. Screenshot `1418_more_after.png` in the improvement report. Data-bearing surfaces (rankings rows, event info links, match detail team stacks) get the same pass once a working API key lands.

gregmarra:
I think this visually isn't worth it right now - we can revisit with future design work.

PR #1435, opened 2026-06-20
docs(tooling): scripts/pr-screenshots.sh + PR screenshot workflow

gregmarra:
## What

Makes "before/after screenshots in the PR description" a repeatable, documented technique.

- **`scripts/pr-screenshots.sh`** — uploads screenshots and rewrites the PR body's screenshot block in place. Images are hosted on an orphan **`screenshot-assets`** branch (own history, never merged) and served from `raw.githubusercontent.com`, so they render inline **without committing image files** to the repo. Built with pure git plumbing, so it never touches your working tree / index / current branch. Filenames are content-hashed to dodge GitHub's image-proxy cache. Works on fork PRs (run as a maintainer).
- **`.claude/CLAUDE.md`** — documents this as the standard way, including pointing the debug build at prod for real data, deep-link navigation, and **how to flush the asset branch** when it gets too full.

## Why not release assets

This org has **immutable releases** enabled, so release assets can't be used as a rewritable image bucket (HTTP 422 on re-upload). The orphan-branch approach is the working alternative.

## Flushing

`git push origin --delete screenshot-assets` resets the branch (the script recreates it). This breaks links in already-merged PRs, so only flush when those are done.

## Proof

Already used to add rendering before/after shots to #1404, #1179, #1405, #1406.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -34,6 +34,53 @@ scripts/emu launch <package/activity>            # force-stop and start activity
 
 Use `find` before `tap` to verify unique matching; use `list` to explore the UI hierarchy.
 
+## PR Screenshots (before/after in PR descriptions)
+
+This is the standard way to put before/after screenshots in a PR — WITHOUT committing
+image files. Images live on an orphan `screenshot-assets` branch (its own history, never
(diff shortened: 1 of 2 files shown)


bherbst (approved):
I don't have access to Claude to test, but looks reasonable to me

PR #1439, opened 2026-06-23
fix(tv): density-bucketed launcher banner + TV-sized icon (#1437)

### LINKED ISSUE #1437: TV: Tile image gets granulated onFocus
When you d-pad onto the tile image for the app, it zooms in and gets pixelated.

YouTube Music also gets pixelated, but other apps like Prime Video do not.

Are we following the correct spec for image size?

gregmarra:
Fixes #1437.

## Problem
D-padding onto the TV app's launcher tile zooms the banner and it granulates/pixelates. The Android TV launcher rasterizes `android:banner` at the asset's **intrinsic** size and then scales that bitmap on focus — so our single banner, a vector with an intrinsic `320dp×180dp` (the legacy xhdpi floor), upscales and softens. Google's own ["Squaring the Circle on Google TV"](https://medium.com/androiddevelopers/squaring-the-circle-on-google-tv-e1ee37fe247e) notes insufficient-resolution assets render "aliased or blurry" in the launcher.

## Fix
Export the vector to the **full density ladder** from the [TV app icon guidelines](https://developer.android.com/design/ui/tv/guides/system/tv-app-icon-guidelines) and point `android:banner` at the mipmaps:

| density | px |
|---|---|
| mdpi | 160×90 |
| hdpi | 240×135 |
| xhdpi | 320×180 |
| xxhdpi | 480×270 |
| xxxhdpi | 640×360 |

This matches what [Jellyfin](https://github.com/jellyfin/jellyfin-androidtv) and other OSS TV clients ship (density-bucketed banners), so the launcher always has a crisp source to scale on focus.

## How it's generated
`scripts/export-tv-banner.py` renders the vector → SVG → PNG with [`resvg`](https://github.com/linebender/resvg) (cross-platform: `brew install resvg` / `scoop install resvg` / `cargo install resvg`; falls back to `rsvg-convert`). It emits both source-set variants:
- **main** (`tv/src/main/res/mipmap-*`) — `#3F51B5` brand field
- **debug** (`tv/src/debug/res/mipmap-*`) — `#770000` red field, preserving the debug tint we previously got from the vector

The vector (`tv/src/main/res/drawable/tv_banner.xml`) stays the **source of truth** — edit it, re-run the script. Kept in the repo (`UnusedResources` ignored) for that reason.

## Test plan
- [x] `:tv:lintDebug` green
- [x] main (blue) + debug (red) banners render correctly at all 5 densities
- [ ] On the TV emulator: confirm the focused tile is crisp (no granulation) vs. before

---

---

## Follow-up: square launcher icon at TV sizes
Per Google's [TV app icon guidelines](https://developer.android.com/design/ui/tv/guides/system/tv-app-icon-guidelines): the legacy `ic_launcher` rasters were phone-sized (xhdpi **96×96**); the spec wants **≥160×160 (xhdpi)** up to **320×320 (xxxhdpi)**. Since `minSdk` is 23, API 23–25 falls back to these (API 26+ uses the adaptive vector icon). Regenerated from the lamp vector at TV sizes via **`scripts/export-tv-icon.py`** — main (blue) + debug (red) — replacing the webps.

Note: full-bleed TV brand icons trip `IconLauncherShape` (a phone-icon rule that also doesn't decode the old webp); disabled for `:tv` with a rationale.

_(An earlier draft also shrank the banner foreground for an "18% safe zone" — reverted: Google's spec gives a numeric safe zone only for the square icon, not the banner, and the banner already had a comfortable ~16% margin.)_

### Test plan
- [x] `:tv:lintDebug` green
- [x] Icon renders crisp at all 5 densities (blue + red)
- [ ] On an actual low-density / older Android TV: confirm the focused banner is crisp (not reproducible on the 4K AVD — it's density-specific)

## Screenshots

**Banner in the Google TV launcher — Favorite Apps row (Television_4K emulator, debug build = red field)**

**Banner detail (2× zoom) — crisp lamp + wordmark at the density the launcher picks**


gregmarra:
## On-device verification (responding to the "blurry assets" note)

Booted the **Television_4K** emulator (3840×2160, density 640 / xxxhdpi) and captured the banner as the **Google TV Launcher actually renders it**:

**Crisp on-device** — the launcher draws the banner card at ~**496×280 px** (measured via uiautomator: `[216,1380][712,1660]`), which is a *downscale* from our 640×360 xxxhdpi asset, so it stays sharp:



**Why it can look blurry in the diff** — the banner card is a fixed ~124 dp at every density, so the matching-density bucket is always *larger* than the rendered size (the launcher never upscales it). What looks blurry is the **raw low-density file viewed larger than its native size** (e.g. the mdpi 160×90 PNG in GitHub's image viewer). Left = raw mdpi upscaled; right = the real render:



**Also confirmed the export path is near-optimal**: resvg renders directly at each density (no downscale step), aspect matches 16:9 (no stretch), and a 4× supersample vs the direct render is visually identical (edge-sharpness within ~5%). So the density-bucket approach here is correct and renders sharp on-device — no change needed.

bherbst (approved):

bherbst:
https://developer.android.com/design/ui/tv/guides/system/tv-app-icon-guidelines#adaptive-banner
indicates we could use an adaptive icon here too, which would support a vector instead of rasterized PNGs

gregmarra:
Something is funky on my physical TV where the onHover zoom in state is getting blurry. I am not sure why - we already have a vector for this asset so I wouldn't expect it. I would vastly prefer to not ship bitmaps!

gregmarra:
I'm going to ship this and see if it fixes it. I don't think this does much harm, even if it is less elegant.

Issue #1392, opened 2026-06-10
Unify loading/empty/error state handling (offline spinner-forever; flash-of-empty on first fetch)


gregmarra:
## Problem

State handling diverges per screen, producing two opposite failure modes:

- **Infinite spinner offline:** `TeamsViewModel.kt:37-47` and `DistrictsViewModel.kt:39-44` map `isEmpty -> Loading` with no "first load finished" escape, so a fresh install with no network spins forever. Their "No teams found"/"No districts found" empty states are unreachable code (`TeamsScreen.kt:81`, `DistrictsScreen.kt:106`).
- **Flash of "No rankings/awards":** event detail tabs render `EmptyBox` the moment Room emits an empty list, while the first network fetch is still in flight (`EventAwardsTab.kt:33-38`, `EventRankingsTab.kt:123-129`, fed by always-non-null flows in `EventDetailViewModel.kt:62-180`).
- **Errors invisible:** `RefreshableViewModel.kt:32-34` logs and swallows refresh failures; `MatchDetailScreen.kt:123-127` shows a spinner forever for a bad/uncached deep link.

`EventsViewModel.kt:45-60` already solves this correctly with a `hasLoaded` flag — the pattern just isn't propagated.

## Proposed approach

Propagate the `hasLoaded` pattern: emit `Empty` only after the first refresh completes; emit `UiState.Error(retry)` when refresh fails and the cache is empty; surface a snackbar when refresh fails but cached data exists; pass `isRefreshing` into the event-tab empty checks so an empty list during first fetch shows `LoadingBox`.

## Why not now

Touches most ViewModels/screens and deserves its own focused PR series with per-screen verification. Surfaced by the PM seat of the improvement-survey panel (2026-06-10).

gregmarra:
Resolved across a focused PR series (all merged) — the three concrete failure modes are fixed and the `hasLoaded`/`RefreshOutcome` pattern from `EventsViewModel` is now propagated:

- **Infinite spinner offline + unreachable Teams/Districts empty states** → #1421
- **Flash of "No rankings/awards" during first fetch** (event-detail tabs) → #1440 (threads `isRefreshing` into the empty checks)
- **Swallowed refresh errors** → #1399 (repositories throw instead of swallowing) + #1441 (a bad/uncached match deep link now shows "Couldn't load match" instead of an endless spinner)

[...]

PR #1441, opened 2026-06-24
fix(matches): no endless spinner on a bad/uncached match deep link

gregmarra:
Part of #1392 (the "errors invisible — `MatchDetailScreen` spins forever for a bad/uncached deep link" failure mode).

## Problem
`MatchDetailScreen` rendered `LoadingBox` whenever `uiState.match == null`, with no escape. So opening a **bad/stale match deep link** (404) — or an **uncached match while offline** — spun **forever**.

## Fix
Track the refresh outcome the same way `TeamsViewModel`/`DistrictsViewModel` do (`refreshingTracked` + `RefreshOutcome`). Once a refresh has *finished* with still no match, surface an `EmptyBox` instead of the spinner. `PullToRefreshBox` already wraps the screen, so pull-to-refresh is the retry.

The message is **"Couldn't load match"** (not "Match not found") because a null match after a failed refresh can mean either a genuine 404 *or* a fetch that failed offline — and I verified it covers both.

## UI verification
On the emulator (debug build → local dev server):
- Bad deep link (`/match/2099fake_qm1`) → was an endless spinner, now resolves to **"Couldn't load match"** once the refresh fails (screenshot below). Gate releases; no infinite spinner.
- A valid match deep link still renders its content (the `match != null` path is unchanged by the diff).

## Screenshots

**Bad/uncached match deep link — now resolves to 'Couldn't load match' (was: endless spinner). Debug build, local server.**

### DIFF
--- a/app/src/main/kotlin/com/thebluealliance/android/ui/matches/MatchDetailScreen.kt
+++ b/app/src/main/kotlin/com/thebluealliance/android/ui/matches/MatchDetailScreen.kt
@@ -46,6 +46,7 @@ import com.thebluealliance.android.domain.model.Match
 import com.thebluealliance.android.domain.model.PlayoffType
 import com.thebluealliance.android.domain.model.displayTitle
 import com.thebluealliance.android.domain.rpBonuses
+import com.thebluealliance.android.ui.common.EmptyBox
 import com.thebluealliance.android.ui.common.LoadingBox
 import com.thebluealliance.android.ui.common.shareTbaUrl
 import com.thebluealliance.android.ui.components.MediaGridItem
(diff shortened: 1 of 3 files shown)


gregmarra:
Hey! Great fix for the endless spinner.

One minor UX improvement idea: when a pull-to-refresh is triggered while the error state is currently showing, `loadFailed` becomes `false` again (since the new request is pending). This transitions the screen from the `EmptyBox` error page back to the full-screen `LoadingBox` (center spinner).

It might be a smoother experience to keep the error screen/message visible during the active pull-to-refresh instead of showing a full-screen layout shift back to the spinner. The spinner at the top of the `PullToRefreshBox` is already sufficient feedback that a refresh is in progress. What do you think?

gregmarra:
Hey this is Claude — appreciate the close read! I went to make this change, then traced it on the actual branch and I think the error screen already stays put during a pull-to-refresh. The key is `RefreshableViewModel.refreshingTracked`: it never sets `outcome` back to `PENDING` — only `SUCCEEDED`/`FAILED` on completion. So after a failed load, `outcome` stays non-`PENDING` through the next refresh, `loadFailed = match == null && outcome != PENDING` stays `true`, and the `EmptyBox` keeps showing (with the `PullToRefreshBox` top spinner). The full-screen `LoadingBox` only renders on the initial `PENDING` load.

I verified on a device build: deep-linked a bad match → "Couldn't load match", then pull-to-refreshed → logcat shows the refresh firing (a 2nd 404) and the error message stays on screen the whole time, no full-screen spinner. So I think this is already handled. Did you maybe catch the spinner via the current `dev.41` build? That one predates this PR and still has the endless-spinner bug it fixes.

bherbst (approved):

PR #1442, opened 2026-06-24
perf(startup): add Baseline Profile to speed up cold start (#1395)

### LINKED ISSUE #1395: Add Baseline Profiles + macrobenchmark module
## Problem

A 100% Compose, Navigation-3, Hilt app shipping with R8 minification has no Baseline Profile and no `profileinstaller` dependency (`gradle/libs.versions.toml` has no baselineprofile/macrobenchmark entries; no `:baselineprofile` module in `settings.gradle.kts`). Cold start and first-scroll run interpreted/JIT after every install and update, relying only on library-shipped profiles and Play cloud profiles, which take days-to-weeks to warm after each release — painful for an app that ships updates around competition season. Typical gains for this app profile are 20-30% on cold start.

## Proposed approach

- Add a `:baselineprofile` macrobenchmark module with the `androidx.baselineprofile` AGP plugin.
- Record a startup + events-list-scroll + event-detail journey (`scripts/emu` shows these flows are already automatable).
- Apply the plugin to `:app` so profiles merge into the AAB; add a CI job to regenerate per release tag.
- Verify with `StartupTimingMetric` before/after.

## Why not now

New Gradle module + CI wiring + benchmark recording is a multi-day chunk. Surfaced by the Android Framework seat of the improvement-survey panel (2026-06-10).

gregmarra:
Closes #1395.

Adds an [AndroidX Baseline Profile](https://developer.android.com/topic/performance/baselineprofiles/overview) so the first cold launch AOT-compiles the hot startup + events-list + event-detail paths instead of relying on JIT (or waiting days for a Play cloud profile).

## What this does
A new `:baselineprofile` module (`com.android.test`) drives the real at-venue journey — cold start → scroll the events list → deep-link a fixed, fully-played event (`2024micmp4`) → page through **Rankings / Matches / Alliances** — and records the classes/methods touched. The captured rules are **committed** to `app/src/main/generated/baselineProfiles/` (41,218 rules; 3,891 app-specific) and merged into the release AAB at build time **with no device in the loop** (`automaticGenerationDuringBuild = false`).

## Measured win
Cold-start time-to-initial-display, 10 cold runs each:

| Compilation mode | Median TTID | Tail (max) |
| --- | --- | --- |
| `None()` — JIT, no profile | 163 ms | 267 ms |
| `Partial(BaselineProfileMode.Require)` — committed profile | **138 ms** | **160 ms** |

**~15% faster median**, and the profile collapses the cold-JIT tail spike (267 → 160 ms). `Require` mode passing also proves the committed profile is actually bundled.

> ⚠️ **Directional only.** These were measured on an **emulator** with `androidx.benchmark.suppressErrors=EMULATOR`. Macrobenchmark refuses emulator runs by default because their timing isn't representative — run `:baselineprofile:connectedBenchmarkReleaseAndroidTest` on a **physical device** for a trustworthy figure (the real-device win is typically larger). The profile *generation* ran on a real connected emulator.

## Supporting wiring
- **testTags → UiAutomator:** `events_list` and `event_rankings` are surfaced as bare resource-ids via `Modifier.semantics { testTagsAsResourceId = true }` on the app root, selected with single-arg `By.res(...)`. The journey's hard-fail guards assert **real content rendered** (`childCount > 0`), not just an always-visible tab label.
- **`ApiKeyProvider`:** a baked `BuildConfig.TBA_API_KEY` (debug / benchmark capture) is authoritative and skips Firebase Remote Config; **shipped releases bake an empty key and keep using Remote Config exactly as before** (no behavior change for users).
- **`StartupBenchmark`** measures `None` vs `Require` for before/after tracking.
- **CI** — `.github/workflows/baseline-profile.yaml` regenerates the profile **on-demand only** (`workflow_dispatch`) and opens a PR. Deliberately **decoupled from release tags** (regenerating on a tag would only land in the *following* release and couple every release to the slowest job in the repo). A slightly stale profile is still a valid win.
- **`baselineprofile/README.md`** documents when/how to regenerate and the measurement caveat.

### One setup step for maintainers
The on-demand CI job needs a repo secret **`BASELINE_PROFILE_TBA_READ_KEY`** — a read-only [TBA APIv3 key](https://www.thebluealliance.com/account) — so the capture journey loads real content (the journey hard-fails on an empty list rather than commit a content-free profile). Local regeneration uses `tba.api.key.benchmark` in `local.properties` instead; see the README.

## Review
Reviewed by an adversarial 5-seat panel (PM, Designer, TL, FRC Nerd, Android Nerd) — all five signed off after the journey was made to assert real rendered content and the CI was decoupled from release tags. Lint (`:app:lintDebug`, warningsAsErrors) is green.


gregmarra:
Going to skip this for now, we can revisit next season.

Issue #1396, opened 2026-06-10
targetSdk 37 (Android 17) readiness punch list


gregmarra:
## Context

`compileSdk` is already 37 everywhere; `targetSdk` is deliberately held at 36 (see comments in the module build files). Android 17 (API 37) is currently in **Beta 4**, so the target bump should wait for stable — but these behavior changes will apply the day we bump, so tracking them now:

## Targeting-37 checklist (from the Android 17 behavior-changes docs)

- [ ] **Certificate Transparency on by default** — already opted in early via `network_security_config.xml` (CT PR in the a17-certificate-transparency branch), so this becomes a no-op. Verify prod + Cloudflare/Let's Encrypt cert chains stay CT-compliant.
- [ ] **`ACCESS_LOCAL_NETWORK` runtime permission** — mandatory enforcement at target 37. Audit whether the debug build's `http://10.0.2.2:8080` dev-server path counts as local network on device builds; production talks only to public HTTPS so the release app should be unaffected. Decide whether the debug manifest needs the permission.
- [ ] **BAL hardening** — verify no `MODE_BACKGROUND_ACTIVITY_START_ALLOWED` usage (none found in-app today); notification `PendingIntent`s should be re-checked against the new `IntentSender` restrictions.
- [ ] **Lock-free `MessageQueue` + unmodifiable `static final` fields** — audit third-party libs for reflection on `MessageQueue` internals or static-final writes (run the app + test suite on an API 37 emulator; watch for `IllegalAccessException`/`UnsatisfiedLinkError`).
- [ ] **Large-screen orientation/resizability opt-outs removed** (sw>=600dp) — we don't opt out today; verify tablet layouts on API 37.
- [ ] **Background audio hardening / ECH** — N/A (no audio playback; ECH needs networking-library support and is opportunistic).
- [ ] **App memory limits (all apps)** — consider logging `ApplicationExitInfo` descriptions containing `MemoryLimiter:AnonSwap` to spot memory-limit kills in the wild.

## Validation done so far (2026-06-10)

- App installs, launches, and renders correctly on the API 37 emulator (`Medium_Phone`, android-37.0 google_apis_playstore_ps16k image).
- Certificate Transparency opt-in landed across `:app`/`:tv`/`:wear`.

## Why not now

Android 17 is beta; the repo policy is to bump targetSdk deliberately on stable. This issue is the punch list for that bump.

gregmarra:
[...]

PR #1444, opened 2026-06-24
fix(tooling)+docs: ACCESS_LOCAL_NETWORK for the Android 17 local-dev backend (#1396)

gregmarra:
Part of #1396 (targetSdk 37 / Android 17 readiness).

## Problem
The `targetSdk 36 → 37` bump (#1422) put the app under Android 17's [Local Network Protections](https://developer.android.com/about/versions/17/behavior-changes-17): reaching an RFC1918 address now requires the **`ACCESS_LOCAL_NETWORK`** runtime permission. The local dev backend is `http://10.0.2.2:8080` (in `10.0.0.0/8`), so the **debug** build is affected. Production talks to public HTTPS — unaffected.

The debug manifest already **declares** the permission, but `ACCESS_LOCAL_NETWORK` is a **dangerous** permission — declaring it doesn't grant it. So a fresh debug install silently can't reach the local server (connections just time out) until someone runs the grant by hand. I hit exactly this while verifying another PR on an Android 17 emulator.

## Fix (two parts)
**1. Tooling.** `scripts/install-and-launch.sh` grants the permission automatically right after `installDebug`, so the local-dev workflow works on Android 17 with no manual step. Best-effort (`|| true`), no-op on pre-17 / when already granted.

**2. Docs.** The auto-grant only helps developers who use that script — Android Studio's ▶ button and a direct `./gradlew :app:installDebug` both bypass it and hit the same silent timeout. So the requirement is now documented where each doc prescribes a bypassing install path:
- **README.md** "Option B: Run the local backend" — the canonical note + the one-time `adb shell pm grant` command.
- **.claude/CLAUDE.md** Build Commands — a one-line pointer (it prescribes `installDebug` + manual relaunch).
- **AGENTS.md** fan-out per-worktree block — the grant inline, so that runnable sequence is correct on Android 17.

## Verified
On a Pixel API 37 emulator: `dumpsys package` showed `ACCESS_LOCAL_NETWORK: granted=false` after a plain install (hence the silent block); `pm grant …` flips it to `granted=true`, restoring local-server reachability. (A doc-surface audit confirmed README is the canonical home and that `screenshot-playbook.sh` / `copilot-setup-steps.yml` are unaffected — they use the release/mock paths.)

The rest of the #1396 punch list (the bump itself, Certificate Transparency, and the other Android 17 behavior changes) is audited on the issue.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -12,12 +12,11 @@ The TBA web server (backend + frontend) is checked out at `~/codez/the-blue-alli
 ./gradlew :app:testDebugUnitTest
 ```
 
-After installing a debug build, always relaunch the app on the emulator:
+After installing a debug build, always relaunch the app on the emulator with `scripts/emu launch`
+(it also grants Android 17's `ACCESS_LOCAL_NETWORK` so the `10.0.2.2` local backend stays reachable):
 ```bash
-adb shell am force-stop com.thebluealliance.androidclient.development && adb shell am start -n com.thebluealliance.androidclient.development/com.thebluealliance.android.MainActivity
+scripts/emu launch com.thebluealliance.androidclient.development/com.thebluealliance.android.MainActivity
 ```
-
-On Android 17 (targetSdk 37), reaching the `10.0.2.2` local backend needs the `ACCESS_LOCAL_NETWORK` runtime permission granted (declaring it isn't enough). `scripts/install-and-launch.sh` does this for you; with a manual `installDebug`, grant it once: `adb shell pm grant com.thebluealliance.androidclient.development android.permission.ACCESS_LOCAL_NETWORK`.
 
 ## Emulator Interaction
 
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -17,6 +17,8 @@ After installing a debug build, always relaunch the app on the emulator:
 adb shell am force-stop com.thebluealliance.androidclient.development && adb shell am start -n com.thebluealliance.androidclient.development/com.thebluealliance.android.MainActivity
 ```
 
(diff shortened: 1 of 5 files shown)


gregmarra:
Hey! While reviewing the changes, I had a thought on how we could simplify the Android 17 `ACCESS_LOCAL_NETWORK` permission grant.

Instead of documenting the manual `adb shell pm grant` command in the README, AGENTS.md, and CLAUDE.md files, we could add the grant logic directly into the `scripts/emu` launch subcommand (which already extracts the package name from the component):

```python
def cmd_launch(args):
    pkg = args.component.split("/")[0]
    # Auto-grant local network permission for Android 17+ emulators (fails silently on pre-17)
    adb("shell", "pm", "grant", pkg, "android.permission.ACCESS_LOCAL_NETWORK")
    adb("shell", "am", "force-stop", pkg)
    adb_check("shell", "am", "start", "-n", args.component)
    print(f"Launched {args.component}")
```

This would ensure that any time `scripts/emu launch` is run (by developers or scripts), it automatically grants the permission without needing any extra setup documentation or manual steps. What do you think?

gregmarra:
Hey this is Claude — great call, that's much cleaner. Done: the grant now lives in `scripts/emu`'s `launch` subcommand (best-effort via `adb()`, so it's a silent no-op on pre-17 / already-granted / not-declared). Everything else routes through it:

- `install-and-launch.sh` now launches via `scripts/emu launch` instead of raw `adb am start`.
- The AGENTS.md fan-out block dropped its explicit grant line (the `scripts/emu launch` it already calls covers it).
- CLAUDE.md's relaunch uses `scripts/emu launch`.
- README keeps just the Android-Studio-▶ case — the one launch path that bypasses both scripts and still needs the one-time manual grant.

Smoke-tested on a live emulator: revoked the permission → `scripts/emu launch` → `granted=true` + app launched.

bherbst (approved):

Issue #1393, opened 2026-06-10
Unify the 3-way forked network stack into :core-network (RFC 0004 follow-up)


gregmarra:
## Problem

RFC 0004 (#1364) moved DTOs into `:core-network` and explicitly deferred unifying the client stack. That fork is drifting:

- `ApiKeyProvider` exists as three near-identical copies (`app/.../config/ApiKeyProvider.kt`, `wear/.../data/ApiKeyProvider.kt`, `tv/.../data/ApiKeyProvider.kt`).
- Json configs diverged: `:app` = `{ignoreUnknownKeys, encodeDefaults}`, `:wear` = `{ignoreUnknownKeys}`, `:tv` = `{ignoreUnknownKeys, coerceInputValues, isLenient}` — a null-for-non-nullable API field crashes `:app`/`:wear` but not `:tv` (`app/di/NetworkModule.kt:33-37`, `wear/data/WearNetworkModule.kt:33-36`, `tv/data/api/TbaApi.kt:17-22`).
- Cache sizes are 20/5/10 MB and timeouts differ for no documented reason.
- The wear Remote Config race (c2f6df2ce) was exactly this class of triplicated-code bug.

## Proposed approach

Next PR in the RFC 0004 series: move `ApiKeyProvider` plus a `TbaClientFactory(baseUrl, apiKey, cacheDir, cacheSize)` — one canonical Json config, API-key interceptor, logging, timeouts — into `:core-network`; each module's DI calls the factory. Pick one Json config deliberately (`ignoreUnknownKeys + coerceInputValues` is the safest superset) and document why.

## Why not now

Medium-size cross-module refactor; should follow the RFC process the repo already uses for this series. Surfaced by the Tech Lead seat of the improvement-survey panel (2026-06-10).

PR #1447, opened 2026-06-25
fix(network): keep null elements in place in the rankings sort_orders arrays (#1445)

### LINKED ISSUE #1445: Harden DTO deserialization against nulls coerceInputValues doesn't cover (follow-up to #1393)
Follow-up to #1393 / #1443 (network-stack unification).

#1393 set the canonical `Json` to `coerceInputValues = true`, which fixes the most common in-event crash (a `null` for a non-nullable field **that has a default** — e.g. an unplayed match's `score`, which falls back to `-1`). But `coerceInputValues` is **narrower than "no null ever crashes"**, and two real gaps remain in `:core-network` DTOs:

### 1. `null` *elements* inside non-null collections — `coerceInputValues` never helps
A `null` element in a `List<NonNull>` always throws, regardless of the flag:
- `RankingItemDto.sortOrders: List<Double>` and `extraStats: List<Double>` — e.g. `"sort_orders": [2.0, null, 0.0]`
- `MatchDto.teamKeys: List<String>`, `surrogateTeamKeys: List<String>`
- `RankingDto.rankings: List<RankingItemDto>`, `AwardDto.recipientList: List<AwardRecipientDto>`

`sort_orders` is the realistic one — TBA can emit a null component for a team with incomplete ranking data.

### 2. Non-null primitives with **no default** — `coerceInputValues` only rescues *defaulted* fields
A `null` for these still throws (no default to coerce to):
- `RankingItemDto.rank: Int`, `RankingSortOrderDto.precision: Int`
- `MatchDto.matchNumber: Int`, `setNumber: Int`
- `AwardDto.awardType: Int`, `AwardDto.year: Int`, `TeamDto.teamNumber: Int`

These are core fields TBA reliably populates, so real-world risk is low — but the codebase shouldn't depend on the server never sending a null for a Kotlin non-null field.

## Proposed approach
- For the collection-element case: either make element types nullable + filter (`List<Double?>` → `filterNotNull()`), or add a small null-dropping serializer for `List<Double>`/`List<String>`.
- For the no-default primitives: give each a sensible default (so `coerceInputValues` catches them) **or** make them nullable where `0`/`-1` would be a misleading sentinel (e.g. `rank`).
- Consider a single shared "lenient list" serializer in `:core-network` so all DTOs get consistent treatment.

## Verification
Add unit tests in `:core-network` that decode payloads with (a) a null inside `sort_orders` and (b) a null `rank`/`awardType` through `TbaClientFactory.json` and assert they degrade gracefully instead of throwing. (Surfaced by the FRC-domain reviewer on #1443.)

gregmarra:
Closes #1445. **Stacked on #1443 (#1393)** — base is `unify-network-stack`; retarget to `main` once #1443 merges.

## Problem
A team with incomplete ranking data can emit a `null` **element** inside the index-aligned `sort_orders` / `extra_stats` / `sort_order_info` arrays. The canonical `Json`'s `coerceInputValues` rescues a `null` for a *defaulted property* but does nothing for a `null` *list element* — it throws and blanks the **whole rankings response**.

## Fix
Keep the `null` **in place** (not dropped) so the columns stay aligned:
- `RankingItemDto.sortOrders` / `extraStats` → `List<Double?>` (the rankings UI renders that one cell `--`).
- `RankingResponseDto.sortOrderInfo` / `extraStatsInfo` → `List<RankingSortOrderDto?>?`; the mapper coalesces a null element to a blank-name placeholder, so the domain/UI type is unchanged and stays index-aligned.

JSON `null` round-trips cleanly through the Room entity (a `NaN` sentinel would not).

## Scope (per review — thanks @bherbst)
Originally this also null-**dropped** several other lists defensively. Per review, a null in `team_keys`/`picks`/etc. isn't a valid backend response (it'd be a bad-ingest bug), so the speculative front-end handling was removed. The PR now covers **only** the rankings arrays — the one place TBA realistically emits a null element. Dropped: `team_keys`/`surrogate_team_keys`, alliance `picks`/`declines`, the rankings-array drop, and the `rank`/`precision` defaults (+ the `NullDroppingListSerializer`).

## Verified
- New `:core-network` test `RankingNullElementTest` (**4/4**) decodes through the real `TbaClientFactory.json`.
- `:app`/`:tv`/`:wear` `compileDebugKotlin` succeed.

### DIFF
--- a/app/src/main/kotlin/com/thebluealliance/android/data/mappers/Mappers.kt
+++ b/app/src/main/kotlin/com/thebluealliance/android/data/mappers/Mappers.kt
@@ -288,9 +288,16 @@ fun RankingResponseDto.toSortOrderEntity(eventKey: String) =
         extraStatsInfo = json.encodeToString(extraStatsInfo ?: emptyList()),
     )
 
+// A null element is kept in place (decoded as RankingSortOrderDto?) and coalesced to a
+// blank-name placeholder, so this column stays index-aligned with its sort_orders values
+// instead of shifting; the UI falls back to "Sort N" for a blank name. #1445.
+private val placeholderSortOrder = RankingSortOrder(name = "", precision = 2)
+
 fun EventRankingSortOrderEntity.getSortOrderInfo(): List<RankingSortOrder> =
     try {
-        json.decodeFromString<List<RankingSortOrderDto>>(sortOrderInfo).map { it.toDomain() }
+        json
+            .decodeFromString<List<RankingSortOrderDto?>>(sortOrderInfo)
+            .map { it?.toDomain() ?: placeholderSortOrder }
     } catch (_: Exception) {
         emptyList()
     }
@@ -298,7 +305,9 @@ fun EventRankingSortOrderEntity.getSortOrderInfo(): List<RankingSortOrder> =
 fun EventRankingSortOrderEntity.getExtraStatsInfo(): List<RankingSortOrder> =
     try {
         if (extraStatsInfo != null) {
-            json.decodeFromString<List<RankingSortOrderDto>>(extraStatsInfo).map { it.toDomain() }
+            json
+                .decodeFromString<List<RankingSortOrderDto?>>(extraStatsInfo)
+                .map { it?.toDomain() ?: placeholderSortOrder }
         } else {
             emptyList()
         }
--- a/app/src/main/kotlin/com/thebluealliance/android/domain/model/Ranking.kt
+++ b/app/src/main/kotlin/com/thebluealliance/android/domain/model/Ranking.kt
@@ -10,6 +10,8 @@ data class Ranking(
     val losses: Int,
     val ties: Int,
     val qualAverage: Double?,
-    val sortOrders: List<Double> = emptyList(),
-    val extraStats: List<Double> = emptyList(),
+    // Nullable elements: a null is a missing component kept in place so it stays index-aligned
+    // with sortOrderInfo (the UI renders that cell blank). #1445.
+    val sortOrders: List<Double?> = emptyList(),
+    val extraStats: List<Double?> = emptyList(),
 )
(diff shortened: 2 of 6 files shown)


core-network/src/test/kotlin/com/thebluealliance/android/data/remote/dto/NullDroppingDeserializationTest.kt
```diff
@@ -0,0 +1,98 @@
...
+            )
+        assertEquals(0, item.rank)
+    }
+
+    @Test
+    fun `null element inside team_keys is dropped`() {
+        val alliance =
+            json.decodeFromString(
+                MatchAllianceDto.serializer(),
+                """{"score":10,"team_keys":["frc1",null,"frc2"]}""",
```
bherbst: This feels to me like it shouldn't be considered a valid response from the backend.
I don't mind band-aiding it on the front-end, but would be interested in improving the response too.

bherbst (approved):

gregmarra:
Hey this is Claude — agreed, a null `team_keys` element isn't a valid response. On the backend `match_converter` just passes `match.alliances[alliance]["teams"]` straight through, so a null there would be a bad-ingest bug, not normal output.

Good call — I've **dropped** the speculative front-end null-handling for the cases that shouldn't happen: `team_keys`/`surrogate_team_keys`, alliance `picks`/`declines`, the rankings-array drop, and the `rank`/`precision` defaults (and removed the `NullDroppingListSerializer`). The PR now keeps **only** the genuinely-realistic case: a null *element* inside the index-aligned `sort_orders`/`extra_stats`/`sort_order_info` arrays (a team with incomplete ranking data), which it keeps in place + renders blank rather than dropping.

Happy to take a follow-up pass at the FMS ingestion parsers if you'd like to stop nulls at the source instead.

PR #1458, opened 2026-06-28
chore: pin the Gradle daemon JVM to 17

gregmarra:
## What

`gradle/gradle-daemon-jvm.properties` pinned the Gradle **daemon JVM to 21**, the lone outlier — everything else is **17**:

| Knob | Was | Now |
|---|---|---|
| Daemon JVM pin | **21** | **17** |
| Compile target (all modules) | 17 | 17 |
| CI (`setup-java`) | 17 | 17 |
| AGP 9.2 min / default JDK | 17 | 17 |

The `21` was an artifact of `updateDaemonJvm` capturing a local JDK 21 — not a requirement (AGP 9.2's minimum *and* default is 17). This pins the build JVM to 17 to match the compile target, CI, and AGP's recommended JDK. We'll bump deliberately when there's a concrete reason.

## How

Regenerated with `./gradlew updateDaemonJvm --jvm-version=17` (via a temporary `foojay-resolver-convention` plugin to provide the toolchain download repo, then reverted), so the daemon auto-download URLs now resolve **JDK 17** rather than 21 — no provisioning regression for contributors who lack a local 17.

## Verification

- `gradle/gradle-daemon-jvm.properties` → `toolchainVersion=17` + JDK-17 foojay URLs (only file changed)
- Daemon comes up on 17 and `:core-network:compileDebugKotlin` builds ✅
- CI's `setup-java` 17 satisfies the new criteria natively


bherbst (approved):
Why _aren't_ we using 21? (I don't think we need to)

gregmarra:
17 is the gradle default, per https://developer.android.com/build/releases/agp-9-2-0-release-notes

PR #1464, opened 2026-06-28
fix(notifications): collapse duplicates for two followed teams in the same match

### LINKED ISSUE #1461: Bug: Duplicate push notifications when following multiple teams in the same match/alliance
## Reported on Chief Delphi

[TBA 2026 site updates & features thread, post #56](https://www.chiefdelphi.com/t/the-blue-alliance-2026-site-updates-features/515744/56) — @waxton:

> if you are following multiple teams on myTBA, and they do something together (play a match, join an elims alliance), you will get 2 notifications for the same thing.

Suggested fix: collapse into one, e.g. `Team 254/1323 CANCMP Q61 Results`.

## Problem

When a user follows multiple teams via myTBA and those teams participate in the same underlying event (the same qual/elim match, the same alliance selection, the same awards ceremony, etc.), TBA's backend fans out one push per subscribed `team_key`. The Android client treats each push independently and posts a separate system notification for each. The user sees N notifications for one real-world event (one per followed team involved), instead of a single collapsed notification.

## Repro

1. On myTBA, follow two teams that are playing in the same match (e.g. 254 and 1323), with "Match score / Final results" notifications enabled.
2. Wait for that match's results to post (or trigger alliance selection involving both teams).
3. Observe two near-identical system notifications arrive for the one match — one keyed to each team.

## Root cause / where it lives

The client posts one system notification per incoming FCM message with no cross-message deduplication, and the notification ID is derived from the full payload (which includes `team_key`), so per-team pushes never collapse onto each other.

- `app/src/main/kotlin/com/thebluealliance/android/messaging/TBAFirebaseMessagingService.kt:68` — `val notificationId = message.data.hashCode()`
- `app/src/main/kotlin/com/thebluealliance/android/messaging/TBAFirebaseMessagingService.kt:69` — `manager.notify(notificationId, notification)`

Because the per-team pushes carry distinct `team_key` values, `message.data` differs between them, so `message.data.hashCode()` differs, so the two `notify()` calls use different IDs and neither replaces the other — they stack.

The builder confirms `team_key` is a first-class part of each message and is even baked into the tap target:
- `app/src/main/kotlin/com/thebluealliance/android/messaging/NotificationBuilder.kt:36-39` — pulls `event_key`, `match_key`, `team_key`, `notification_type` from `message.data`.
- `app/src/main/kotlin/com/thebluealliance/android/messaging/NotificationBuilder.kt:62-63` — deep-link URI includes `$teamKey`.
- `app/src/main/kotlin/com/thebluealliance/android/messaging/NotificationBuilder.kt:70-71` — PendingIntent `requestCode` includes `teamKey`.

There is currently **no** grouping, tag, summary, or dedup of any kind: a grep of `app/src/main` for `setGroup` / `GroupSummary` / `InboxStyle` / `.tag` returns nothing, and there is no persisted "seen notifications" store. The only `notify()` calls for incoming pushes are the two lines above (the three `notify(9001/9002/9003, …)` calls in `ui/settings/SettingsScreen.kt:172/188/204` are local test-notification buttons, unrelated).

**Confidence: high** that this is the right code path for *how* the duplicate is displayed. **Caveat:** I cannot see the backend payload from this repo, so the exact set of fields that differ between the two pushes (and whether the backend could instead send a single multi-team push) is inferred from the field names the client reads. The client-side fix below is robust regardless of which fields differ, because it keys off the shared `match_key`/`event_key`/`notification_type` rather than the per-team `team_key`.

## Proposed approach

Collapse pushes that describe the same underlying event so the second (and later) push updates the first notification rather than adding a new one. Two complementary options, fitting the existing `messaging/` structure:

1. **Dedup the notification ID off the shared event, not the whole payload (minimal, recommended first step).** In `TBAFirebaseMessagingService.kt:68`, replace `message.data.hashCode()` with an ID built from the fields that are *common* across teams for one event — e.g. `listOf(notification_type, match_key, event_key).joinToString("|").hashCode()` (deliberately excluding `team_key`). Per-team pushes for the same match then share one ID, so the later push replaces the earlier one. Mirror the same exclusion in `NotificationBuilder`'s `requestCode` (NotificationBuilder.kt:70-71) and deep-link URI (NotificationBuilder.kt:62-63) so the surviving notification's tap target stays valid. This already removes the "2 notifications" symptom (you'd see one, for whichever team's push landed last).

2. **Merge the team labels into one body (delivers @waxton's exact ask).** To show `Team 254/1323 … Results` rather than just the last team, the builder needs to remember which teams it has already shown for a given (`notification_type`,`match_key`/`event_key`) key within a short window and rebuild the content text accordingly. A small in-memory/`DataStore`-backed map keyed on the collapsed ID, holding the set of `team_key`s seen, is enough; on each push, add the new team, re-render the title/body listing all teams, and re-`notify()` with the shared ID. This logic belongs in `NotificationBuilder` (it already owns title/body construction) or a thin helper it calls, keeping `TBAFirebaseMessagingService` as the dumb dispatcher it is today.

Optionally use Android notification **groups** (`setGroup` + a group-summary notification) as a presentation layer on top, but groups alone won't merge two same-match items into one line — the ID/body collapse above is what addresses the report.

Suggested scope for a first PR: implement option 1 (ID collapse + matching requestCode/URI) plus a unit test for the messaging path (none exists today under `app/src/test`), then follow up with option 2 for the `Team A/B` body.

gregmarra:
## What

Closes #1461. Two followed teams in the **same match** produced **two duplicate notifications**; this collapses them to one.

## Root cause

`TBAFirebaseMessagingService` posted with `notificationId = message.data.hashCode()`, which includes `team_key`. The backend fans out **one push per subscribed team**, so two followed teams in the same match got two different ids → two stacked notifications.

## Fix

`NotificationBuilder.collapseId(data)` keys the id off the **shared `(notification_type, match_key)`** for match notifications, so per-team pushes share one id and the later push replaces the earlier — **one notification per match**. Notifications without a match context (e.g. alliance selection) keep a per-message id, so nothing over-collapses.

## Visual evidence

A debug-only **"Test: Duplicate match (2 followed teams)"** button (Settings → Debug) posts two same-match / different-team pushes through the real notification path. Before/after below: **two notifications → one**.

## Verification

- Unit test `NotificationCollapseTest` — same match + different team → same id; different match → different id; different type → different id; non-match → per-message.
- `:app:lintDebug` ✅ (warningsAsErrors) and `:app:testDebugUnitTest` ✅
- On-device (phone emulator): harness fires two pushes → BEFORE two notifications, AFTER one.

## Follow-up

Merging the team labels into one body ("Team 254/1323 … Q61 Results", @waxton's exact ask) is a larger stateful change — left as a follow-up; this PR removes the duplicate-notification symptom.

## Screenshots

| | Before | After |
|---|---|---|
| **Same-match notifications (2 followed teams)** |  |  |


gregmarra:
✅ **5-seat panel: all approve, no majors.**

Folded in the one worthwhile minor — **[Android] `setOnlyAlertOnce(true)`** so the collapsing second push updates the notification *silently* instead of re-buzzing for content already shown.

Confirmed non-issues / scoped follow-ups:
- **Tap target** — Android + PM verified the surviving notification still lands on the correct `MatchDetail` (routing is by `match_key`, team-independent), so collapsing loses nothing.
- **Event-level types** (alliance_selection, schedule_updated, …) can still double for users following 2+ teams at one event — out of scope here; collapsing them needs per-type granularity (e.g. `comp_level`), noted in code + as a follow-up.

bherbst (approved):

PR #1466, opened 2026-06-28
fix(mytba): render + remove event_team (model_type 3) subscriptions

### LINKED ISSUE #1460: Bug: myTBA Notifications tab can't show/remove team@event (model_type 3) subscriptions
## Reported on Chief Delphi

https://www.chiefdelphi.com/t/the-blue-alliance-2026-site-updates-features/515744/31 (also /33, /35)

@waxton: *"apparently a while ago I set up a notification for 2471 at a specific event in mytba, and now I can't seem to get rid of it"* — and when asked whether it shows up in the in-app myTBA list: *"not seeing anything there unfortunately."* So a myTBA subscription exists server-side and still fires FCM notifications, but it does not appear in the app's myTBA Notifications tab, leaving no way to delete it from the app.

## Problem

A server-side myTBA subscription for a **team at a specific event** (e.g. team 2471 @ event `XXXX`) keeps firing notifications but does not surface in the in-app **myTBA → Notifications** list, so the user can't remove it.

This is a "team-at-event" subscription. On the TBA backend the myTBA model-type enum is `event=0, team=1, match=2, event_team=3, district=4`. A "team 2471 at a specific event" subscription is `model_type = 3` (event_team), with a `model_key` like `2024xxxx_frc2471`. The Android client only models three of these types:

`app/src/main/kotlin/com/thebluealliance/android/domain/model/ModelType.kt`
```kotlin
object ModelType {
    const val EVENT = 0
    const val TEAM = 1
    const val MATCH = 2
}
```
There is no `EVENT_TEAM` (3) or `DISTRICT` (4). The new Compose app also offers no UI to *create* a team@event subscription, so any such subscription a user holds was created by the legacy app / website and can only be managed there.

## What the code actually does with a model_type=3 subscription

Tracing the data path, nothing filters type 3 out of the DB or the list:

- DTO keeps the raw int: `SubscriptionDto.modelType` — `core-network/src/main/kotlin/com/thebluealliance/android/data/remote/dto/ClientApiDtos.kt:29-33`
- Repository stores it verbatim (no type filter): `MyTBARepository.refreshSubscriptions()` — `app/src/main/kotlin/com/thebluealliance/android/data/repository/MyTBARepository.kt:90-105`, and re-emits it: `observeSubscriptions()` — `MyTBARepository.kt:48-57`
- DAO returns all rows: `SubscriptionDao.observeAll()` — `app/src/main/kotlin/com/thebluealliance/android/data/local/dao/SubscriptionDao.kt:12-13`
- The list renders every subscription, falling back to `"Other"` for unknown types: `NotificationsTab` / `SubscriptionItem` — `app/src/main/kotlin/com/thebluealliance/android/ui/mytba/MyTBAScreen.kt:466-481` and `492-497` (`else -> "Other"`)

So a type-3 row *should* appear, but degraded: its display name can't resolve (the name resolver only handles TEAM/EVENT/MATCH — `MyTBAViewModel.kt:54-62`), so it shows the raw key (`MyTBAScreen.kt:470`, `displayNames[...] ?: subscription.modelKey`), and tapping it does nothing (the `when` has no type-3 branch — `MyTBAScreen.kt:472-476`). The user likely either (a) didn't recognize the raw-key "Other" row as their subscription, or (b) never saw it because the list was empty on their device.

The "empty list" path is plausible and is the second half of this bug: the Notifications tab is populated **only** from local Room, which is filled by `refreshSubscriptions()`. That refresh runs only on a sign-in *transition* (`MyTBAViewModel.init`, `MyTBAViewModel.kt:84-104`) or via pull-to-refresh (`MyTBAScreen.kt:252`). There is no refresh on screen open, so a user who signed in long ago can see a stale/empty Notifications tab until they manually pull-to-refresh.

Confidence: medium. I'm confident the model-type gap is real and that `ModelType` not modeling `event_team=3` is the structural cause of the degraded/confusing row (verified in source). I could not run the app against the user's account to confirm whether their specific row renders-as-"Other" vs. is-absent-due-to-stale-cache; both are reachable from the code above.

## Repro

1. Have a myTBA `event_team` (model_type 3) subscription on your account — e.g. created via the website: subscribe team 2471 at a specific event. (The new Android app cannot create one, so use the site.)
2. Open the Android app, sign in, go to **myTBA → Notifications**.
3. Pull to refresh.
4. Observe: at best, a row whose title is the raw key (e.g. `2024xxxx_frc2471`) labeled "Other" that does nothing when tapped; at worst (stale cache, no refresh triggered) no row at all. Notifications for it keep arriving, but there is no clean way to identify/remove it.

## Root cause / where it lives

- Primary: `ModelType` does not model `event_team` (3) (nor `district` 4) — `app/src/main/kotlin/com/thebluealliance/android/domain/model/ModelType.kt:1-6`. Downstream display-name resolution (`MyTBAViewModel.kt:54-62`), the type label (`MyTBAScreen.kt:492-497` → "Other"), and click navigation (`MyTBAScreen.kt:472-476`, no branch) all therefore degrade for type 3.
- Secondary: the Notifications list is only as fresh as the last `refreshSubscriptions()`, which is only triggered on a sign-in transition or pull-to-refresh (`MyTBAViewModel.kt:84-104`, `MyTBAScreen.kt:252`) — there's no refresh-on-open, so a long-signed-in user can see an empty list.

Note: removal itself is type-agnostic and already works — `removeSubscription` round-trips `modelKey`/`modelType` through `updatePreferences` (`MyTBAViewModel.kt:122-142`, `MyTBARepository.kt:161-191`). So once a type-3 row is shown with a working Remove action, the user can delete it.

## Proposed approach

1. Add the missing model types so they're first-class:
   - In `ModelType`, add `const val EVENT_TEAM = 3` (and `DISTRICT = 4` for completeness) — `domain/model/ModelType.kt`.
2. Make the Notifications/Favorites rows recognizable and actionable for `event_team`:
   - Resolve a friendly display name. An `event_team` key is `{eventKey}_{teamKey}` (e.g. `2024xxxx_frc2471`); in `buildMyTBADisplayNames` (`MyTBAViewModel.kt:177-198`) and the key-collection logic (`MyTBAViewModel.kt:54-62`), split the key, observe both the team and the event, and format e.g. `"2471 @ 2024 <Event Name>"`. Feed the split team key into `teamRepository.observeTeams(...)` and the event key into `eventRepository.observeEvents(...)`.
   - Give it a real type label instead of "Other" in `SubscriptionItem`/`FavoriteItem` (`MyTBAScreen.kt:378-384`, `491-497`), e.g. "Team @ Event".
   - Add a click branch (e.g. navigate to the event, or the team's page) in the `when` blocks (`MyTBAScreen.kt:336-340`, `472-476`), or make it non-clickable but clearly labeled. The existing Remove action already works via `removeSubscription`, so even a non-navigable row solves the user's core complaint (deleting it).
3. Guarantee freshness so the row actually shows up: trigger `refreshSubscriptions()` (and favorites) when the myTBA screen becomes visible, not only on the sign-in transition — e.g. an explicit refresh in `MyTBAViewModel.init`/on first subscription, or a `LaunchedEffect` in `MyTBAScreen` calling `viewModel.refresh()`. This avoids relying on the user discovering pull-to-refresh.

This keeps raw model-type ints at the DTO/DB edge (consistent with the existing pattern) while surfacing the enum-typed concept in the UI, matching the project's "no magic ints / establish enums for domain codes" convention.

gregmarra:
## What

Closes #1460. myTBA can hold **event_team** subscriptions (a team @ a specific event), created by the old app / website. The new app couldn't surface them — `ModelType` only knew EVENT/TEAM/MATCH, so they rendered as **"Other"** with the **raw key** (`2024micmp4_frc2471`) and weren't recognizable, leaving users unable to tell what they were to remove them ([Chief Delphi](https://www.chiefdelphi.com/t/the-blue-alliance-2026-site-updates-features/515744/31)).

## Fix

- `ModelType` gains `EVENT_TEAM` (3) and `DISTRICT` (4).
- The myTBA ViewModel resolves the composite `{eventKey}_{teamKey}` key to a friendly **"team @ event"** name (observes both halves; falls back to the parsed key parts when the team/event aren't cached).
- `SubscriptionItem` / `FavoriteItem` give them a real type label ("Team @ Event" / "District") instead of "Other"; tapping an event_team row opens its event.
- Removal is already type-agnostic, so the orphaned subscription is now both **identifiable and removable** (the ⋮ → Remove).

## Visual evidence

Signed in locally via the Firebase Auth emulator (no real Google account — see #1465) and seeded an `event_team` subscription with a **debug-only** harness button. Before → after below.

## Verification

- Unit tests — `MyTBADisplayNamesTest`: event_team resolves to "team @ event", and falls back to parsed key parts when uncached.
- `:app:lintDebug` ✅ (warningsAsErrors) and `:app:testDebugUnitTest` ✅
- On-device (phone emulator, signed in): **before** = raw key + "Other"; **after** = "2471 @ 2024micmp4" / "Team @ Event", removable via ⋮.

## Follow-up

Refresh-on-screen-open (so a long-signed-in user's list isn't stale) is a related secondary fix left out of scope here.

## Screenshots

| | Before | After |
|---|---|---|
| **myTBA → Notifications (orphaned event_team subscription)** |  |  |

### DIFF
--- a/app/src/main/kotlin/com/thebluealliance/android/domain/model/ModelType.kt
+++ b/app/src/main/kotlin/com/thebluealliance/android/domain/model/ModelType.kt
@@ -4,4 +4,9 @@ object ModelType {
     const val EVENT = 0
     const val TEAM = 1
     const val MATCH = 2
+
+    // myTBA also models these; the old app / website can create them even though the new app
+    // doesn't yet, so they must render + be removable here (#1460).
+    const val EVENT_TEAM = 3 // key is "{eventKey}_{teamKey}", e.g. 2024micmp4_frc2471
+    const val DISTRICT = 4
 }
--- a/app/src/main/kotlin/com/thebluealliance/android/ui/mytba/MyTBAScreen.kt
+++ b/app/src/main/kotlin/com/thebluealliance/android/ui/mytba/MyTBAScreen.kt
@@ -337,6 +337,8 @@ private fun FavoritesTab(
                         ModelType.TEAM -> onNavigateToTeam(favorite.modelKey)
                         ModelType.EVENT -> onNavigateToEvent(favorite.modelKey)
                         ModelType.MATCH -> onNavigateToMatch(favorite.modelKey)
+                        ModelType.EVENT_TEAM ->
+                            onNavigateToEvent(favorite.modelKey.substringBefore('_'))
                     }
                 },
                 canPinShortcuts = canPinShortcuts,
@@ -380,6 +382,8 @@ private fun FavoriteItem(
             ModelType.EVENT -> "Event"
             ModelType.TEAM -> "Team"
             ModelType.MATCH -> "Match"
+            ModelType.EVENT_TEAM -> "Team @ Event"
+            ModelType.DISTRICT -> "District"
             else -> "Other"
         }
     var menuExpanded by remember { mutableStateOf(false) }
@@ -473,6 +477,8 @@ private fun NotificationsTab(
                         ModelType.TEAM -> onNavigateToTeam(subscription.modelKey)
                         ModelType.EVENT -> onNavigateToEvent(subscription.modelKey)
                         ModelType.MATCH -> onNavigateToMatch(subscription.modelKey)
+                        ModelType.EVENT_TEAM ->
+                            onNavigateToEvent(subscription.modelKey.substringBefore('_'))
                     }
                 },
                 onRemove = { onRemoveSubscription(subscription) },
@@ -493,6 +499,8 @@ private fun SubscriptionItem(
             ModelType.EVENT -> "Event"
             ModelType.TEAM -> "Team"
             ModelType.MATCH -> "Match"
+            ModelType.EVENT_TEAM -> "Team @ Event"
+            ModelType.DISTRICT -> "District"
             else -> "Other"
         }
     var menuExpanded by remember { mutableStateOf(false) }
--- a/app/src/main/kotlin/com/thebluealliance/android/ui/mytba/MyTBAViewModel.kt
+++ b/app/src/main/kotlin/com/thebluealliance/android/ui/mytba/MyTBAViewModel.kt
@@ -53,13 +53,28 @@ class MyTBAViewModel
             }.distinctUntilChanged().flatMapLatest { keyedTypes ->
                 fun keysOf(type: Int) = keyedTypes.filter { it.second == type }.map { it.first }
                 val matchKeys = keysOf(ModelType.MATCH)
+                // event_team keys are "{eventKey}_{teamKey}" — observe both halves so the row
+                // resolves to a friendly "team @ event" name instead of the raw key (#1460).
+                val eventTeamKeys = keysOf(ModelType.EVENT_TEAM)
+                val teamKeys =
+                    (
+                        keysOf(
+                            ModelType.TEAM,
+                        ) + eventTeamKeys.map { it.substringAfter('_') }
+                    ).distinct()
                 val eventKeys =
-                    (keysOf(ModelType.EVENT) + matchKeys.map { it.substringBefore('_') }).distinct()
+                    (
+                        keysOf(ModelType.EVENT) +
+                            matchKeys.map { it.substringBefore('_') } +
+                            eventTeamKeys.map { it.substringBefore('_') }
+                    ).distinct()
                 combine(
-                    teamRepository.observeTeams(keysOf(ModelType.TEAM)),
+                    teamRepository.observeTeams(teamKeys),
                     eventRepository.observeEvents(eventKeys),
                     matchRepository.observeMatches(matchKeys),
-                ) { teams, events, matches -> buildMyTBADisplayNames(teams, events, matches) }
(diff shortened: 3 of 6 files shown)


gregmarra:
✅ **5-seat panel: all approve** (TL / PM / Designer / Android / FRC), no majors.

**Scope — view + remove only, no creation.** This renders + removes *existing* event_team subscriptions; it adds **no way to create** them. The event/team detail screens still create only plain EVENT/TEAM subs, and the sole insert path is the **debug-only** seed harness (`if (BuildConfig.DEBUG)`, stripped from release) used purely to reproduce the bug for the screenshot.

Panel minors (non-blocking, noted as follow-ups):
- **DISTRICT (type 4)** is label-only — it gets a "District" label (better than "Other") but no friendly-name resolution or tap target yet; intentionally outside #1460's event_team scope.
- The "after" screenshot shows the uncached fallback name (`2471 @ 2024micmp4`); the friendly `team @ year Event` path is unit-tested.

bherbst (approved):

PR #1484, opened 2026-07-20
fix(wear): only honor team_number config intent in debug builds

gregmarra:
## What

`TeamTrackingComplicationConfigActivity` is `exported=true` (required by the complication `PROVIDER_CONFIG` contract). The `onCreate` path accepted a `team_number` intent extra and persisted it in **all build types**, with **no user confirmation** and **no validation** — so any installed app or an ADB command could silently repoint the tracked team. Because the value bypassed the digit filter the on-screen field applies, a non-numeric string could be stored and interpolated straight into API paths (`frc$team`).

## Fix

- Gate the intent path on `BuildConfig.DEBUG` so it only exists as an ADB/testing affordance; R8 strips the branch from release builds entirely.
- Route both the intent path and the on-screen field through a single `normalizeTeamNumber()` helper, so the intent can't bypass the field's digit validation.
- **Manifest intentionally unchanged.** `exported=true` + `category.DEFAULT` is the platform-mandated shape for a complication config activity (the system picker launches it via an implicit intent, which requires `DEFAULT`). Hardening the *code* to treat the intent as untrusted is the correct layer.

## Evidence

- `:wear:assembleDebug` + `:wear:lintDebug` both green.
- Debug affordance preserved: `am start … -e team_number 254` still sets the team; the tracker renders team 254 (screenshot below).
- Release behavior: the `team_number` branch is compiled out, so a launched activity only shows the config UI, which requires a physical Save tap.

Reviewed for design correctness by a Fable subagent (approach + manifest reasoning verified against the complication data-source platform contract).

## Screenshots

**Debug intent path preserved — tracker renders team 254 after `am start -e team_number 254`**

### DIFF
--- a/wear/src/main/kotlin/com/thebluealliance/android/wear/complication/TeamTrackingComplicationConfigActivity.kt
+++ b/wear/src/main/kotlin/com/thebluealliance/android/wear/complication/TeamTrackingComplicationConfigActivity.kt
@@ -31,6 +31,7 @@ import androidx.wear.compose.material3.MaterialTheme
 import androidx.wear.compose.material3.ScreenScaffold
 import androidx.wear.compose.material3.Text
 import androidx.wear.watchface.complications.datasource.ComplicationDataSourceUpdateRequester
+import com.thebluealliance.android.wear.BuildConfig
 import com.thebluealliance.android.wear.R
 import com.thebluealliance.android.wear.tracker.TeamTrackerPreferences
 import com.thebluealliance.android.wear.ui.TbaWearTheme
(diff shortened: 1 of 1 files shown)


gregmarra (commented):
Code logic looks great! Hardening the exported `TeamTrackingComplicationConfigActivity` behind `BuildConfig.DEBUG` and normalizing team numbers to digits only is the right security posture. It looks like GitHub Actions CI matrix hasn't executed for this PR branch yet — once CI is re-triggered and green, this is ready to merge.

gregmarra (commented):
Approved / LGTM! CI completed cleanly with all 8 checks green. Ready to merge.

PR #1486, opened 2026-07-20
refactor(wear): remove dead teamNickname data flow + its per-refresh network call

gregmarra:
## What

`TeamTrackingComplicationWorker.updateAppTracker` made a dedicated `api.getTeam(teamKey)` network call on **every refresh** solely to populate `prefs.teamNickname`. That value was copied into `TeamTrackerState`, but **no composable ever reads it** — `TeamHeader` renders only the team number. So the field, the pref, and the extra network round-trip produced nothing visible (and on a watch, that round-trip is often proxied over Bluetooth — real battery cost).

## Fix (delete, not render)

Removed the whole dead flow:
- Worker: the `prefs.teamNickname = try { api.getTeam(...) … }` block (`teamKey` stays — still used for alliance-membership checks).
- `TeamTrackerActivity`: `teamNickname` from `loadState` and from `TeamTrackerState`.
- `TeamTrackerPreferences`: the `teamNickname` var + `KEY_TEAM_NICKNAME`.
- `WearTbaApi`: the now-unused `getTeam()` method + `TeamDto` import.

**Delete vs render:** the finding flagged these as mutually exclusive. The phone app's `TeamTrackingWidget` *does* render `"$teamNumber — $teamNickname"`, so showing it on the watch is plausible — but that's a deliberate UX feature (subtitle placement/truncation on a round display, and it would compete with the record/event subtitle already there), not something to ship silently under a cleanup. Deleting is the conservative, reversible win; a follow-up feature PR can restore it using the widget as a template if desired.

Orphaned `team_nickname` pref on existing installs is harmless (never read) and self-heals via `clearCachedData()`'s full `clear()` on the next reconfigure.

## Evidence

- `:wear:assembleDebug` + `:wear:lintDebug` green.
- Tracker renders identically (it never showed the nickname) — screenshot below.
- Fable subagent reviewed the delete-vs-render fork and the deletion's completeness (grep-clean, `TeamDto` still used by the phone app, no migration needed); approved delete.

## Screenshots

**Tracker renders identically — nickname was never displayed**

### DIFF
--- a/wear/src/main/kotlin/com/thebluealliance/android/wear/data/WearTbaApi.kt
+++ b/wear/src/main/kotlin/com/thebluealliance/android/wear/data/WearTbaApi.kt
@@ -3,7 +3,6 @@ package com.thebluealliance.android.wear.data
 import com.thebluealliance.android.data.remote.dto.EventDto
 import com.thebluealliance.android.data.remote.dto.MatchDto
 import com.thebluealliance.android.data.remote.dto.MediaDto
-import com.thebluealliance.android.data.remote.dto.TeamDto
 import retrofit2.http.GET
 import retrofit2.http.Path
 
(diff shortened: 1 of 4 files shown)


gregmarra (commented):
Approved / LGTM! Great cleanup of the unused `teamNickname` field, state, and preference handling. Removing the unused `getTeam` API call on every refresh worker run is a great saver for data and battery.

PR #1487, opened 2026-07-20
refactor(wear): model alliance as an enum, extract Alliance.of() helper

gregmarra:
## What

The "which alliance is the tracked team on" expression was **copy-pasted 3×** in `TeamTrackingComplicationWorker`, and raw `"red"`/`"blue"` string literals were compared at **~12 sites** across the worker and `TeamTrackerActivity` — contradicting the repo's convention of establishing enums for domain codes and keeping raw codes only at the API/DB edge. The avatar-background `when()` also silently defaulted **any unexpected value to blue**.

## Fix

- New `Alliance` enum (`RED`/`BLUE`) with:
  - `fromKey(raw): Alliance?` — parse a raw TBA/pref value; null for blank/unknown.
  - `of(match, teamKey): Alliance?` — the single extracted helper; null if the team is on neither alliance.
- **Worker** routes all comparisons through the enum; prefs still store the raw `"red"`/`"blue"` string (the persistence edge stays raw). `extractBonusRp` now takes `Alliance?`.
- **State** parses the raw pref once via `fromKey` at the read edge; composables compare `== Alliance.RED/BLUE`.
- **Correctness fix:** an unknown alliance now paints a neutral avatar backdrop instead of blue — so a team **not currently at an event** (a common case: the worker clears match data, leaving alliance unknown) is no longer misrepresented as "on blue".

Behavior is otherwise unchanged: `Alliance.of` is non-null for a team's own matches (`teamMatches` is pre-filtered to matches the team plays), so persisted values and win/loss/tie tallies are identical.

**Follow-ups (out of scope):** hoist `Alliance` to `core-network` to also cover the phone app's ~12 raw red/blue sites; unit-test the enum + record loop (see the wear-tests PR).

## Evidence

- `:wear:assembleDebug` + `:wear:lintDebug` green; all raw `"red"`/`"blue"` literals now confined to `Alliance.kt`.
- Tracker renders correctly (screenshot below). The visual delta (neutral vs blue backdrop) only appears with an avatar present but no current match.
- A Fable subagent produced a full record-loop/persistence/bonus-RP equivalence proof and confirmed the neutral-background change is correct and legible on Wear.

## Screenshots

**Tracker renders correctly after the enum refactor**

### DIFF
--- a/wear/src/main/kotlin/com/thebluealliance/android/wear/tracker/Alliance.kt
+++ b/wear/src/main/kotlin/com/thebluealliance/android/wear/tracker/Alliance.kt
@@ -1,0 +1,34 @@
+package com.thebluealliance.android.wear.tracker
+
+import com.thebluealliance.android.data.remote.dto.MatchDto
+
+/**
+ * Which alliance a team competes on in a match.
+ *
+ * TBA encodes this as the raw strings "red"/"blue" at the API and SharedPreferences edges;
+ * [key] is that wire/storage form. Everything above those edges works with the enum so the
+ * codebase never compares bare "red"/"blue" literals.
+ */
+enum class Alliance(
+    val key: String,
+) {
+    RED("red"),
+    BLUE("blue"),
+    ;
+
+    companion object {
+        /** Parse a raw TBA/pref value; returns null for blank or anything unrecognized. */
+        fun fromKey(raw: String?): Alliance? = entries.firstOrNull { it.key == raw }
+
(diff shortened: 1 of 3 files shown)


gregmarra (commented):
Approved / LGTM! Strong domain modeling improvement replacing magic `red` / `blue` strings with the `Alliance` enum and `Alliance.of()` helper, plus replacing arbitrary color fallback with a neutral background.

PR #1488, opened 2026-07-20
feat(wear): distinct error state with retry on failed tracker refresh

gregmarra:
## What

`TeamTrackerState` tracked `isLoading` but had **no error field**, and `doWork` cleared `isLoading` in its `finally` even when it returned `Result.retry()` after an exception. So a newly-set team whose first fetch **failed** showed only the header + "No upcoming events" — indistinguishable from a team that genuinely has no events, and with no way to retry.

## Fix

- New `lastRefreshFailed` pref; the worker sets it `false` right before `Result.success()` and `true` in `doWork`'s `catch` (the fetch — `fetchTeamEventData` — is deliberately outside the inner try/catch blocks, so its failure reaches the outer catch).
- `TeamTrackerState.lastRefreshFailed` drives a distinct **"Couldn't refresh" + Retry** state, shown only when a refresh failed **and** there's no cached data to fall back on. Cached (stale) data still takes precedence over the error.
- `retryRefresh()` clears the flag, shows the spinner, and re-enqueues an immediate refresh. `clearCachedData()`'s full `clear()` resets the flag on team change so a new team never inherits a stale error.

## Evidence

Before/after below, captured on the Wear emulator. (The debug build's placeholder API key makes the team-254 fetch fail with 401 — a real instance of the exact bug: the old UI reports "No upcoming events" for what is actually a failed fetch.)

- `:wear:assembleDebug` + `:wear:lintDebug` green.
- A Fable subagent verified the flag is set for the right failure (fetch throw → outer catch), the `showError` gating never hides cached data, `clearCachedData` resets it on team change, and the mid-list Retry button doesn't compete with the bottom Change-Team EdgeButton. Noted follow-up (out of scope): an offline retry leaves the spinner up until the network-constrained worker can run — consistent with the screen's existing team-change loading model.

## Screenshots

| | Before | After |
|---|---|---|
| **Failed first fetch (team 254, 401)** |  |  |

### DIFF
--- a/wear/src/main/kotlin/com/thebluealliance/android/wear/tracker/TeamTrackerActivity.kt
+++ b/wear/src/main/kotlin/com/thebluealliance/android/wear/tracker/TeamTrackerActivity.kt
@@ -86,6 +86,7 @@ class TeamTrackerActivity : ComponentActivity() {
                     TeamTrackerScreen(
                         state = currentState,
                         onChangeTeam = { launchConfigActivity() },
+                        onRetry = { retryRefresh() },
                     )
                 }
             }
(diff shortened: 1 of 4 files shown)


gregmarra (commented):
Approved / LGTM! Clear UX improvement adding `lastRefreshFailed` tracking and rendering a retry button when refresh fails without cached data.

PR #1489, opened 2026-07-20
perf(wear): batch the worker's per-refresh SharedPreferences writes

gregmarra:
## What

Every `TeamTrackerPreferences` / `TeamTrackingComplicationPreferences` setter did its own `edit{}.apply()`. So a single `updateAppTracker` fired **~20 async disk writes**, and because the tracker registers an `OnSharedPreferenceChangeListener`, **~20 `loadState()` + recomposition passes** in the foreground — some reading torn intermediate state (e.g. red score updated, blue not yet).

## Fix

Add a `beginBatch()` / `applyBatch()` pair backed by a shared `Editor`. While a batch is open, each setter routes into it via a small `put {}` helper instead of committing on its own; `applyBatch()` commits once. `updateAppTracker` and `updateComplication` wrap their writes in one batch (the suspend nickname fetch is hoisted out of the batch first). `clearCachedData()`/`clear()` keep their own single commits.

Result per refresh: **one atomic commit** — a single disk write, no half-written reads, and one coalesced reload + recomposition instead of ~20.

`doWork`'s `finally` calls `applyBatch()` before `isLoading = false`, so a batch left open by a mid-refresh exception is flushed and the loading flag still commits (never swallowed into a dead editor).

## Evidence

- `:wear:assembleDebug` + `:wear:lintDebug` green; tracker renders with no regression (screenshot below).
- A Fable subagent verified: all 28 setters route correctly and write identical values; exception paths are safe (the tracker's partial batch commits late in `finally` — behavior parity with the old incremental applies — and `isLoading=false` always lands; per-complication partials drop cleanly on discarded, file-isolated instances); no mid-batch torn reads are possible; the single-writer assumption holds (each `doWork` builds its own wrapper instances). It corrected an overstated code comment (Android fires the listener per changed key, so the real wins are disk writes, atomicity, and coalesced recomposition) — the comment was reworded accordingly.

_Note: the batched write path only runs on a successful fetch, which needs a real API key; on-device here the debug key 401s before `updateAppTracker`, so the screenshot verifies no render/crash regression rather than exercising the batch._

## Screenshots

**Tracker renders with no regression after batching the prefs writes**

### DIFF
--- a/wear/src/main/kotlin/com/thebluealliance/android/wear/complication/TeamTrackingComplicationPreferences.kt
+++ b/wear/src/main/kotlin/com/thebluealliance/android/wear/complication/TeamTrackingComplicationPreferences.kt
@@ -16,29 +16,49 @@ class TeamTrackingComplicationPreferences(
     private val prefs: SharedPreferences =
         context.getSharedPreferences("complication_$complicationId", Context.MODE_PRIVATE)
 
+    /** Non-null while a [beginBatch] batch is open; setters write here instead of committing. */
+    private var batchEditor: SharedPreferences.Editor? = null
+
+    /** Route a write to the open [beginBatch] batch, or commit it immediately when none is active. */
+    private inline fun put(write: SharedPreferences.Editor.() -> Unit) {
+        val editor = batchEditor
+        if (editor != null) editor.write() else prefs.edit { write() }
+    }
+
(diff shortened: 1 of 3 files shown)


gregmarra (commented):
Approved / LGTM! Great performance optimization batching `SharedPreferences` writes into single transactions per worker refresh, avoiding multiple disk writes and UI listener churn.

PR #1490, opened 2026-07-20
fix(wear): bound avatar decoding, catch Throwable, cache the alpha mask

gregmarra:
## What

Two problems on the avatar path:

1. **Unbounded decode + `catch(Exception)`.** Both decode sites — `AvatarConverter` (complication) and the tracker's `TeamHeader` — called `BitmapFactory.decodeByteArray(bytes, 0, size)` with no size bounds and caught only `Exception`. An `OutOfMemoryError` (a subclass of `Error`) from a malformed/oversized PNG escapes and can **crash the complication service or the activity**.
2. **Per-request pixel scan.** `AvatarConverter.toAlphaMask` ran a full decode + two-pass per-pixel luminance scan on **every** `onComplicationRequest`, re-deriving the identical mask and allocating two `IntArray`s + a `Bitmap` each render.

## Fix

- New `decodeBoundedBitmap()`: decodes bounds first (`inJustDecodeBounds`), downsamples an oversized image toward `MAX_AVATAR_PX` (96) via a power-of-two `inSampleSize`, and catches `Throwable` (returns null, logs). Both decode sites now use it — the tracker's inline unbounded decode is gone.
- `toMonochromaticImage` memoizes a **single entry keyed by the source base64**, so the decode + luminance scan runs once per distinct avatar instead of once per request. (Only one team is tracked, so all complication instances converge on one key.)

A normal 40×40 FRC avatar decodes at `inSampleSize = 1` → byte-identical output; only oversized images are downsampled.

## Evidence

- `:wear:assembleDebug` + `:wear:lintDebug` green; tracker renders with no regression (screenshot below).
- A Fable subagent verified: `calculateInSampleSize` returns 1 for 40×40 (no visual change), bounds a 4000×4000 to ~125px, and degrades safely on a failed bounds decode (`-1` dims → sampleSize 1 → null); catching `Throwable` in this non-suspend decode is safe (no `CancellationException` concern); the `AtomicReference` memo has only benign races and never returns a wrong-key image; and `toAlphaMask` output is byte-for-byte unchanged.

_Screenshot: team 254's avatar, fetched from prod and decoded through the new bounded/`Throwable`-guarded path, rendering in the tracker header._

## Screenshots

**Real avatar (team 254) decoded via the hardened decodeBoundedBitmap and rendered in TeamHeader**

### DIFF
--- a/wear/src/main/kotlin/com/thebluealliance/android/wear/complication/AvatarConverter.kt
+++ b/wear/src/main/kotlin/com/thebluealliance/android/wear/complication/AvatarConverter.kt
@@ -5,8 +5,10 @@ import android.graphics.BitmapFactory
 import android.graphics.Color
 import android.graphics.drawable.Icon
 import android.util.Base64
+import android.util.Log
 import androidx.core.graphics.createBitmap
 import androidx.wear.watchface.complications.data.MonochromaticImage
+import java.util.concurrent.atomic.AtomicReference
 
 /**
  * Converts a color avatar bitmap (base64-encoded PNG) into an alpha mask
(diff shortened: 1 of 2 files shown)


gregmarra (commented):
Approved / LGTM! Hardening avatar decoding with sample size bounds, caching alpha masks, and catching `Throwable` is essential memory protection on Wear OS devices.

PR #1492, opened 2026-07-20
refactor(wear): drop unused teamNumber param from updateComplication

gregmarra:
## What

`TeamTrackingComplicationWorker.updateComplication` declared a `teamNumber: String` parameter that its body never referenced — it reads only `prefs` and the already-team-scoped `TeamEventData`. Team identity is baked into the fetched data upstream (`doWork` builds `teamKey = "frc$trackerTeam"` before fetching), so the parameter was residual plumbing, likely left over from before the `TeamEventData` extraction.

Removed the parameter and the `trackerTeam` argument at the single call site. No behavior change.

## Evidence

- `:wear:assembleDebug` + `:wear:lintDebug` green; tracker renders unchanged (screenshot below).
- A Fable subagent confirmed the parameter was genuinely unused, the single call site maps correctly by position, and no information is lost (the sibling `updateAppTracker` correctly keeps its own `teamNumber`, which it does use).

## Screenshots

**Tracker renders unchanged after dropping the unused parameter**

### DIFF
--- a/wear/src/main/kotlin/com/thebluealliance/android/wear/worker/TeamTrackingComplicationWorker.kt
+++ b/wear/src/main/kotlin/com/thebluealliance/android/wear/worker/TeamTrackingComplicationWorker.kt
@@ -133,7 +133,7 @@ class TeamTrackingComplicationWorker
                                     applicationContext,
                                     complicationId,
                                 )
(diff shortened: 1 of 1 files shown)


gregmarra (commented):
Approved / LGTM! Clean removal of the unused `teamNumber` parameter from `updateComplication`.

PR #1525, opened 2026-09-01
feat(webcasts): open Twitch webcasts in the Twitch app

gregmarra:
## Motivation

Tapping a Twitch webcast on the event Info tab handed `https://twitch.tv/<channel>` straight to the browser. **Verified on a real Quest 3:** the browser's "open in the Twitch app" interstitial *drops the channel path* — you get dumped on the Twitch home screen, not the stream you asked for. Firing `twitch://stream/<channel>` instead opens the Twitch app directly on the correct channel. The same interstitial runs on phones with Twitch installed, so phone users get the fix too.

## What changed

- When a webcast is a Twitch channel, try `ACTION_VIEW twitch://stream/<channel>` first.
- If nothing handles it (no Twitch app installed → `ActivityNotFoundException`), fall back to the existing https browser behavior, unchanged.
- Non-Twitch webcasts (YouTube, Livestream, unknown types) are completely unchanged.

`startActivity()` does not require package visibility to launch another app's activity ([docs](https://developer.android.com/training/package-visibility/use-cases)), so catching `ActivityNotFoundException` is sufficient — **no `<queries>` entry and no `QUERY_ALL_PACKAGES` needed** on API 30+.

This is a per-type special case at the link-out site, not a new webcast framework. `webcastUrl` moved out of `EventInfoTab.kt` into `util/WebcastLinks.kt` next to the new `webcastAppUri` so both are pure and unit-testable; the label formatting stays in the tab.

## Testing

- New `WebcastLinksTest` (7 tests) covering URL building for each type and the Twitch deep-link choice, including blank-channel and non-Twitch cases.
- `:app:testDebugUnitTest` — pass.
- `:app:ktlintCheck` — pass.
- `:app:lintDebug` — only the 2 pre-existing `AndroidGradlePluginVersion` errors already on `main`; zero new findings.
- On-device Twitch behavior verified on a Quest 3 (see Motivation).

`:tv` (which has its own `WebcastLauncher` for the TV fallback chain) is untouched.

### DIFF
--- a/app/src/main/kotlin/com/thebluealliance/android/ui/events/detail/tabs/EventInfoTab.kt
+++ b/app/src/main/kotlin/com/thebluealliance/android/ui/events/detail/tabs/EventInfoTab.kt
@@ -33,6 +33,8 @@ import com.thebluealliance.android.ui.common.LoadingBox
 import com.thebluealliance.android.ui.components.formatEventDateRange
 import com.thebluealliance.android.ui.events.weekLabel
 import com.thebluealliance.android.util.openUrl
+import com.thebluealliance.android.util.openWebcast
+import com.thebluealliance.android.util.webcastUrl
 
 @Composable
 fun EventInfoTab(
@@ -225,7 +227,7 @@ fun EventInfoTab(
                             Modifier
                                 .padding(top = 4.dp)
                                 .clickable {
-                                    context.openUrl(url)
+                                    context.openWebcast(webcast, url)
                                 },
                         verticalAlignment = Alignment.CenterVertically,
                     ) {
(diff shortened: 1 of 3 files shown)


bherbst (commented):
Is this actually broken? Twitch links seem to be correctly opening the twitch app for me, and that matches my expectations given that the Twitch app declares support for `https://twitch.tv` URLs in their manifest

bherbst:
I can believe the Quest is doing something non-standard here, but that seems like a bug on Quest's end

gregmarra:
It's some combination of bugs where:
1. Clicking the link in TBA should catch in the Twitch Quest app, but instead catches in the Quest Browser
2. The Quest Browser offers to open it in Twitch, but the deeplink doesn't load to the channel, just to the main Twitch activity.

I'll report this bug internally at Meta, but may end up with some some hacky exceptions in the Quest build flavor.

PR #1526, opened 2026-09-02
fix: inline text-link hover padding

gregmarra:
## Problem

A pixel-verified hover audit found a cluster of hand-rolled text links built as
`Modifier.clickable { }` on a wrap-content `Text`, with no padding inside the clickable and no
`clip`. On a mouse/trackpad (Android 17 desktop windowing, ChromeOS, emulator with a pointer) the
hover highlight and the press ripple are drawn over the clickable's own bounds — so they trace the
glyph box exactly, producing a tight rectangle glued to the text instead of a deliberate target.
The same bounds are the touch target, so several of these were also well under 48dp.

## Mechanism / fix

The repo already has the correct idiom in `EventRow.kt`: size first, `clickable` next, `padding`
last — the padding is *inside* the clickable, so it becomes part of the hover/ripple/touch surface.
For elements that do not span a full row, the shape has to be clipped before `clickable` or the
indication paints a hard-edged rectangle.

Two shapes of fix, applied per context:

- **Block-level links** (event name on Match Detail, "Learn more about OPR" on Team@Event Stats):
  `fillMaxWidth()` → `clickable` → `padding`, i.e. a row-style hover band, with the horizontal
  inset moved from the parent into the clickable so the text does not move.
- **Inline links and table cells** (More footer links, Match Detail alliance team numbers,
  Team@Event top-bar title halves, alliance pick slots, Regional Advancement point cells, team
  numbers nested inside already-clickable rows): `clip(shapes.small)` → `clickable` → `padding`,
  so hover reads as a rounded chip around the label. Where a cell is already `weight(...)`-sized,
  only vertical padding is added — the weight already gives it the full column width.

Nested navigation is preserved everywhere. Where a team number sits inside an already-clickable
row (Rankings, Advancement), the inner target keeps its own clipped, padded surface so the two
hover states read as intentionally nested rather than as one band bleeding into another.

Wherever possible, the padding added *inside* the clickable is taken back out of the enclosing
container, so the rest state is unchanged:

- `EventRankingsTab` ranking row: row padding 12dp → 8dp vertical, team cell gains 4dp vertical →
  row height and every child's position are identical.
- `EventAdvancementTab` row: row padding 10dp → 8dp vertical, team-number target gains 2dp → same
  row height.
- `EventAlliancesTab`: the alliance block's 16dp horizontal padding is split 8dp on the block +
  8dp on each child, and `FlowRow`'s 16dp arrangement gap is now supplied by the two slots'
  8dp insets — content stays on the 16dp margin, gaps between picks stay 16dp, and the wrap
  threshold is arithmetically unchanged.
- `MoreScreen` footer: row padding 16dp → 8dp and the separator's surrounding spaces (`" · "` →
  `"·"`) are replaced by the links' own 8dp insets, so "About" still starts on the 16dp margin
  and the gap around the dot is about the same.
- `TeamEventDetailScreen` title: the spaces around `" @ "` are likewise replaced by the two
  halves' 6dp insets.

## Rest-state changes (disclosed)

These could not be fully compensated:

- **Match Detail, event name**: text sits 2dp lower; the time line below it 4dp lower; the block
  is 4dp taller.
- **Match Detail, alliance team numbers**: each team number's target grows from 26dp to 32dp tall
  (+6dp per team, ~18dp per alliance column) and gains a 12dp horizontal inset. Text stays
  horizontally centered, so it does not move sideways.
- **Team@Event top bar**: the title starts ~6dp further right (the team-number link's leading
  inset). Bar height is unchanged.
- **Team@Event Stats, "Learn more about OPR"**: 4dp taller; text position unchanged.
- **More footer**: the version/build line sits ~4dp lower.
- **Regional Advancement rows**: the two event-point cells become 32dp tall. Rows whose
  advancement badge already wraps to two lines are unchanged; rows with a short badge grow by up
  to 8dp. Rows never shrink.

Before/after hover screenshots for this cluster will come in a later consolidated screenshot pass
across all the hover PRs.

## Verification

`:app:assembleDebug`, `:app:testDebugUnitTest`, `:app:lintDebug` all run locally. Lint reports only
the two pre-existing `AndroidGradlePluginVersion` errors that are already failing on clean `main`
(an AGP bump PR is in flight); no new findings.

## Screenshots

| | Before | After |
|---|---|---|
| **More screen link hover** |  |  |

### DIFF
--- a/app/src/main/kotlin/com/thebluealliance/android/ui/events/detail/tabs/EventAdvancementTab.kt
+++ b/app/src/main/kotlin/com/thebluealliance/android/ui/events/detail/tabs/EventAdvancementTab.kt
@@ -33,6 +33,7 @@ import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
 import androidx.compose.ui.draw.rotate
 import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.dp
@@ -129,7 +130,9 @@ private fun AdvancementPointsItem(
                 Modifier
                     .fillMaxWidth()
                     .clickable { expanded = !expanded }
-                    .padding(horizontal = 16.dp, vertical = 10.dp),
+                    // 8.dp here + 2.dp inside the team-number tap target below keeps the
+                    // row's overall height unchanged while the nested target grows.
+                    .padding(horizontal = 16.dp, vertical = 8.dp),
             verticalAlignment = Alignment.CenterVertically,
         ) {
             Text(
@@ -142,7 +145,12 @@ private fun AdvancementPointsItem(
                     text = points.teamKey.teamNumber,
                     style = MaterialTheme.typography.titleMedium,
                     color = MaterialTheme.colorScheme.primary,
-                    modifier = Modifier.clickable { onTeamClick(points.teamKey) },
+                    modifier =
+                        Modifier
+                            .fillMaxWidth()
+                            .clip(MaterialTheme.shapes.small)
+                            .clickable { onTeamClick(points.teamKey) }
+                            .padding(vertical = 2.dp),
                 )
                 if (teamName != null) {
                     Text(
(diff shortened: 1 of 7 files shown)


bherbst (approved):
I think the other fix for these would be to wrap the `Text` components in a `Box` and put the clickable on the `Box` with padding on the `Text` but this solution is probably about the same at the end of the day so doesn't super matter.

Could also consider extracting this combination of modifies to an extension function (or a new Modifier implementation) for reusability since quite a few changes are doing the same thing now

